PR #5, opened 2019-11-21
OM-235 | Added configurations for CI pipelines

klempine:
Changed the Docker configs based on what was configured in the
kukkuu-ui application. These should now include a nginx config
for the health check endpoints. Added a yaml file for Gitlab.

The values in the configs are mostly guesses. We'll see how this
plays out in the actual CI build.


.env
```diff
@@ -1,3 +1,13 @@
+REACT_APP_API_URI=
```
jimppa: Unless this has to do with the CI-pipeline, we are not using this. `REACT_APP_PROFILE_GRAPHQL` is used for specifying the URL to profile-api(graphql).

.env
```diff
@@ -1,3 +1,13 @@
...
+REACT_APP_OIDC_CLIENT_ID="https://api.hel.fi/auth/omahelsinki"
 REACT_APP_PROFILE_AUDIENCE="https://api.hel.fi/auth/profiles"
+REACT_APP_PROFILE_GRAPHQL=
 REACT_APP_OIDC_SCOPE="openid profile $REACT_APP_PROFILE_AUDIENCE"
+REACT_APP_SENTRY_DSN=
+REACT_APP_VERSION=$npm_package_version
+#TRANSLATION_LANGUAGES=en,fi,sv
+#TRANSLATIONS_URL=https://docs.google.com/spreadsheets/d/18dpV7aA-t_6zMRUO5hP7DMfx0q4hdu-AlLpRjkB_0zI/gviz/
+#TRANSLATION_PROJECT_NAME=
+#SASS_PATH=./src/assets/styles
```
jimppa: We are not using SASS for omahessu, so this could be deleted.

.env
klempine: Removed references to API_URI

.env
klempine: Removed it

jimppa (approved):

PR #16, opened 2019-12-30
OM - 262 | Edit profile form

tuovinensanttu:
Editing profile is now possible.


src/profile/components/editProfile/EditProfile.tsx
```diff
@@ -0,0 +1,125 @@
...
+                postalCode: formValues.postalCode,
+                primary: true,
+                addressType: AddressType.OTHER,
+              }
+            : null,
+        ],
+        updateAddresses: [
+          profileData?.myProfile?.primaryAddress?.id
+            ? {
+                id: profileData?.myProfile?.primaryAddress?.id,
```
jimppa: `id` is guaranteed to be something here because of the ternary check so it can be accessed directly without the optional chaining stuff `profileData.myProfile.primaryAddress.id`

jimppa (changes requested):
Check the comment I left regarding optional chaining.

src/profile/components/editProfile/EditProfile.tsx
tuovinensanttu: Fixed.

jimppa (approved):

PR #18, opened 2020-01-08
OM - 264 | Remove profile

tuovinensanttu:


src/common/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div className={classNames(styles.container, props.className)}>
+      <input
+        className={styles.checkbox}
+        type="checkbox"
+        onChange={props.onChange}
+      />
+      <span className={styles.label}>
+        {!props.useTransComponent && props.label}
+        {props.useTransComponent && (
```
jimppa: Could `label` include `useTransComponent`, `components` and `i18Key`? So that the type would be `label: string | React.ReactNode` and usage `<Checkbox label="foo" />` or `<Checkbox label={<Trans i18Key... />}`

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const handleProfileDelete = () => {
+    setDeleteConfirmationModal(false);
+
+    const variables = {
+      id: props.profileID,
+    };
+
+    deleteProfile({ variables })
+      .then(result => {
```
jimppa: Is this `.then()`-block needed?

src/profile/helpers/getServiceNames.ts
```diff
@@ -0,0 +1,13 @@
+import { ServiceConnectionQuery } from '../graphql/__generated__/ServiceConnectionQuery';
+
+export const getServiceTypes = (data?: ServiceConnectionQuery) => {
```
jimppa: I think it would be better if the export would have the same name as the module, or even better a default export! I would change this to `export default function getServiceNames(data?: ServiceConnectionQuery) { ... }`

src/profile/components/profileInformation/ProfileInformation.tsx
```diff
@@ -20,42 +21,45 @@ function ProfileInformation(props: Props) {
...
+              />
+              <LabeledValue
+                label={t('profileInformation.phone')}
+                value={data.myProfile?.primaryPhone?.phone}
+              />
+            </>
+          )}
+        </div>
+      </section>
+      <DeleteProfile profileID={data.myProfile?.id || ''} />
```
jimppa: Would typescript be happy with `{data.myProfile?.id && <DeleteProfile profileID={data.myProfile.id} />`? I don't think the component should be shown without it.

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (result.data) {
+        }
+      })
+      .catch(err => {
+        err.graphQLErrors &&
```
jimppa: I think this part should be refactored into a separate helper or something. Don't like the hard-coded `CANNOT_DELETE_PROFILE_WHILE_SERVICE_CONNECTED_ERROR`, I think it should at least be a constant somewhere.

jimppa (changes requested):
Left some comments on things I think could be improved.

src/common/checkbox/Checkbox.tsx
tuovinensanttu: This approach is much better. Changed.

src/profile/components/deleteProfile/DeleteProfile.tsx
tuovinensanttu: It wasn't needed, removed.

src/profile/helpers/getServiceNames.ts
tuovinensanttu: Changed export to default as suggested. I kept the original name (getServiceTypes) and renamed file to match that.

src/profile/components/profileInformation/ProfileInformation.tsx
tuovinensanttu: Yep, it should't be show without id.

src/profile/components/deleteProfile/DeleteProfile.tsx
tuovinensanttu: Created constant and exported error check to own file.

src/common/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+function Checkbox(props: Props) {
+  return (
+    <div className={classNames(styles.container, props.className)}>
+      <input
+        className={styles.checkbox}
+        type="checkbox"
+        onChange={props.onChange}
+      />
+      <span className={styles.label}>{props.label}</span>
```
jimppa: I missed this the first time, sorry! This `span` should be a `label` associated with the input.

src/profile/components/profileInformation/ProfileInformation.tsx
```diff
@@ -20,42 +21,47 @@ function ProfileInformation(props: Props) {
...
+              <LabeledValue
+                label={t('profileInformation.phone')}
+                value={data.myProfile?.primaryPhone?.phone}
+              />
+            </>
+          )}
+        </div>
+      </section>
+      {data.myProfile?.id && (
+        <DeleteProfile profileID={data.myProfile?.id || ''} />
```
jimppa: Is the conditional still needed in profileID? I would think that typescript would be fine with `profileID={data.myProfile.id}`

jimppa (changes requested):
A few more things I noticed, otherwise looks good!

src/common/checkbox/Checkbox.tsx
tuovinensanttu: Span changed to label & added name prop as required.

src/profile/components/profileInformation/ProfileInformation.tsx
tuovinensanttu: Condition was not needed, removed.

jimppa (approved):

PR #21, opened 2020-01-09
Improve readme

jimppa:
Attempt at improving the readme


klempine (approved):
This is better than before, kudos!

PR #23, opened 2020-01-16
OM - 310 | Add terms of service

tuovinensanttu:


src/tos/components/termsOfService/TermsOfService.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const termsOfService = tosKeys;
+
+  const handleGoBack = () => {
+    history.goBack();
+  };
+
+  return (
+    <PageLayout className={styles.background}>
+      <div className={styles.wrapper}>
```
jimppa: This seems to work fine without the styles from `.wrapper`, so maybe remove `.wrapper` from CSS.

src/i18n/sv.json
```diff
@@ -71,5 +71,36 @@
...
+  },
+  "footer": {
+    "copyright": "Upphovsrätt {{year}}",
+    "reserveRights": "Alla rättigheter förbehållna",
+    "privacy": "Integritetspolicy",
+    "accessibility": "Tillgänglighetsdokument",
+    "terms": "Användarvillkor"
+  },
+  "tos": {
+    "title": "Helsinkiprofiilin käyttöä koskevat käyttöehdot",
```
jimppa: It's a bit strange that title2 has the text 1. XXX. Could you rename some of these titles and messages?
E.g. something like this:
```
"pageTitle": "Helsinkiprofiilin käyttöä koskevat käyttöehdot",
...
...
"termsTitle": "Helsinki profiili - avain kaikkiin kaupungin palveluihin",
"termsText": "Luomalla Helsinkiprofiilin voit...",
"terms1Title": "1. Soveltamisala",
"terms1Terms": "Helsinkiprofiili on Helsingin kaupungin..."
```

src/tos/helpers/getTransComponents.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ReactNode } from 'react';
+
+export default function getTransComponents(links: Array<string> | undefined) {
+  if (links === undefined) return [];
+  const components: Array<ReactNode> = [];
+
+  links.forEach(link => {
```
jimppa: It would be nicer to use `links.map(...)` instead of forEach.

src/tos/helpers/getTransComponents.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ReactNode } from 'react';
+
+export default function getTransComponents(links: Array<string> | undefined) {
```
jimppa: You could get rid of the `if (links ...` by setting a default value of empty-array to links: `getTransComponents(links: Array<string> = []) {`

src/tos/constants/tosConstants.ts
```diff
@@ -0,0 +1,7 @@
+export default {
```
jimppa, line 2: Would this file need a TODO-comment?

jimppa (changes requested):

src/tos/components/termsOfService/TermsOfService.tsx
tuovinensanttu: Removed

src/i18n/sv.json
tuovinensanttu: Changed all the keys to match title numbers.

src/tos/helpers/getTransComponents.tsx
tuovinensanttu: Changed to .map

src/tos/helpers/getTransComponents.tsx
tuovinensanttu: Added default

src/tos/constants/tosConstants.ts
tuovinensanttu: Added simple todo note

jimppa: Forgot to push? :)

jimppa (approved):

PR #27, opened 2020-01-27
OM-362 no service connections

jimppa:
Show text when user does not have any service connections.
Put all graphql ts-types into one generated file.


src/i18n/fi.json
```diff
@@ -14,7 +14,8 @@
   },
   "serviceConnections": {
     "title": "Yhdistetyt palvelut",
-    "explanation": "Olet yhdistänyt seuraavat palvelut tunnukseesi. Voit käyttää samalla profiilillasi useita helsingin kaupungin palveluita sekä tiettyjä kolmansien osapuolten palveluita."
+    "explanation": "Olet yhdistänyt seuraavat palvelut tunnukseesi. Voit käyttää samalla profiilillasi useita helsingin kaupungin palveluita sekä tiettyjä kolmansien osapuolten palveluita.",
+    "empty": "Sinulla ei tällä hetkellä ole palveluita liitettynä profiiliisi."
```
tuovinensanttu: With the current combination of "explanation" + "empty" the message can be little confusing. I think you can remove sentence `Olet yhdistänyt seuraavat palvelut tunnukseesi` to make it more understandable for the user. 

Same for the other languages.

tuovinensanttu (changes requested):

tuovinensanttu (approved):

PR #31, opened 2020-02-06
OM - 450 | Polish connected services

tuovinensanttu:
Includes
- OM - 451
- OM - 453

Added missing heading above used data list.
Prevented HKI_MY_DATA from rendering.
Added service created text.


Jokinen (approved):
Cod looked clean. Checked the UI and it behaved as described 👍Nice job!

PR #32, opened 2020-02-06
OM - 367 | Handle errors in the backend call

tuovinensanttu:
- Added NotificationComponent for showing simple error
- Added error handling to every query and mutation.

### DIFF
--- a/src/common/notification/NotificationComponent.module.css
+++ b/src/common/notification/NotificationComponent.module.css
@@ -1,0 +1,13 @@
+.notification {
+    width: 300px;
+    position: fixed;
+    top: 100px;
+    right: 0;
+}
+
+.notification p {
+    margin: 0;
+}
+
+
+
--- a/src/auth/authenticate.ts
+++ b/src/auth/authenticate.ts
@@ -1,5 +1,9 @@
 import userManager from './userManager';
+import store from '../redux/store';
+import { apiError } from './redux';
 
 export default function(): void {
-  userManager.signinRedirect();
+  userManager.signinRedirect().catch(error => {
+    store.dispatch(apiError(error.toString()));
+  });
 }
--- a/src/auth/components/login/Login.tsx
+++ b/src/auth/components/login/Login.tsx
@@ -1,16 +1,24 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { connect } from 'react-redux';
 
+import { RootState } from '../../../redux/rootReducer';
+import { AuthState, resetApiError } from '../../redux';
 import { ReactComponent as HelsinkiLogo } from '../../../common/svg/HelsinkiLogo.svg';
 import styles from './Login.module.css';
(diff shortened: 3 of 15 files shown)


src/common/notification/NotificationComponent.module.css
```diff
@@ -0,0 +1,13 @@
+.notification {
+    width: 300px;
+    position: fixed;
+    top: 100px;
+    right: 0;
+}
+
+.notification p {
```
Jokinen, line 8: Is this a sort of hack to style a `HDS` defined element?

Jokinen (changes requested):
In the original ticket CORS is specifically mentioned. As I know how to cause this error (by setting the port to something else than `3000`, e.g. `PORT=3001 yarn start` :D) it's the one I tested first. I was able to see an error in the log, but I did not get a notification.

I think the problem here may be that when you get an error like CORS, the request fails all together. It may be that currently you are only handling cases where the request-response trade is successful and the request is returned with an error code. 

The case where the request isn't event sent successfully (because the server is blocking it, because the server is down and can't receive it etc. a network error) will still fail silently.

Might be that it's something else, but the above is a common thing that people miss. I think that this is something that should be addressed before this can be said to fix OM-367. This might help: https://www.apollographql.com/docs/link/links/error/.

Something I don't want to request changes on, but what I want to mention is the way in which the notification functionality has been implemented. The notification element seems to be largely the same everywhere. Wouldn't it be better to declare it once somewhere higher up in the component tree and then use redux/context to open it when the application faces an error? This approach has its own pitfalls, but for a very simple case like open ->  show something generic -> close there shouldn't be that many things that need scaffolding.

One other comment I want to point out is that the notification appeared in a sort of weird place for me. I was only able to test it by setting the notification component visible manually, so I don't know if that's the reason.

Anyway, for me the notification was pressed right against the right side of the page while still having some top margin on for instance this page, e.g.
Screenshot 2020-02-06 at 14 27 54

src/common/notification/NotificationComponent.module.css
tuovinensanttu: For some reason hds Notification component doesn't have className prop. So only way to restrict it's size is to use some sort of wrapper. I'm not sure if that counts as a hack.

tuovinensanttu:
The CORS error was caused by test tunnistamo, error is now caught and notification should pop up. I tested apollo hooks with no internet connection (should be somewhat similar to CORS error) and notification is shown correctly. Ill keep provided link in mind if in the future we want to provide more detailed errors to users.

Regarding the other stuff. 
- This project (and Youth membership) are build with mentality "use redux as little as possible" and that's why I went with this approach. 
- Positioning of the notification may not be final, but for now it's in "right" position. That's about the same position used in Varaamo & kukkuu notifications.

Jokinen (approved):
A CORS error now displayed a notification on the page for me 👍 LGTM

PR #33, opened 2020-02-06
OM - 364 | Handle no connected services

tuovinensanttu:
Modal will show different message based on amount of connected services.


Jokinen (approved):
Code is clean, worked when tested in UI 👍

PR #35, opened 2020-02-20
OM - 491 | Logout functionality

tuovinensanttu:
Added logic that allows user to log out. Because the design included additions to header & dropdown functionality, I added those as well.

Redirecting back to application after logout isn't working, that's a feature in current version of Tunnistamo.


Jokinen (approved):
Looks to be working nicely 👍

We should get this component into HDS :D

PR #37, opened 2020-03-04
OM - 516 | Logout after profile deletion

tuovinensanttu:
After profile is successfully deleted, user is redirected to new page. This page was added mainly because tunnistamo's logout is quite awful.


Jokinen (approved):
Looks good 👍Tested in UI and I was redirected correctly.

PR #39, opened 2020-03-23
Add production stage config

tuomas777:
Changed also OIDC client ID to match the naming convention used in KuVa Tunnistamo.


tuovinensanttu (approved):
Looks good to me 👍

PR #40, opened 2020-03-25
OM -553 | Fixed UI breaking with long names.

tuovinensanttu:
- Name shown in header is reduced to first name only and about 17 characters.
- Name in pageHeading is also reduced to one line.

Backend url was changed today (26.3), enviroment variable change is included in this PR.


.env.development
```diff
@@ -1,3 +1,3 @@
 REACT_APP_OIDC_AUTHORITY="https://tunnistamo.test.kuva.hel.ninja/"
 REACT_APP_OIDC_CLIENT_ID="https://api.hel.fi/auth/omahelsinki"
-REACT_APP_PROFILE_GRAPHQL="https://helsinkiprofile.test.kuva.hel.ninja/graphql/"
+REACT_APP_PROFILE_GRAPHQL="https://profiili-api.test.kuva.hel.ninja//graphql/"
```
Jokinen: The double `/` here is probably accidental.

Jokinen (approved):
Code looks good  👍Tested the UI and it worked.

PR #43, opened 2020-03-30
OM - 563 | Add profile language to forms

tuovinensanttu:
- Added profile language to create and edit forms.
- Added button to edit view which navigates back to profile view.


src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -27,17 +30,19 @@ export type FormValues = {
...
+  setEditing: () => void;
   isSubmitting: boolean;
   profile: FormValues;
   onValues: (values: FormValues) => void;
 };
 
 function EditProfileForm(props: Props) {
   const { t } = useTranslation();
-
+  const languages = ['FINNISH', 'ENGLISH', 'SWEDISH'];
```
Jokinen: You declare these languages in at least two places in this change. You could change these hard coded values into constants etc. Minor detail.

Jokinen (approved):
Code looked good 👍 I checked the UI and everything was working as intended.

One thing I noticed is that the select element uses a different styling to that of the other inputs. As I recommended using the native select in a different PR in a different project I feel partly responsible :D 

In case you didn't know, it's actually possible to style the select even when using the native element. Your styling won't be applied unless you explicitly declare that you don't want the default appearance.

I created a quick example PR that shows the idea behind the approach: https://github.com/City-of-Helsinki/open-city-profile-ui/pull/46

I used this link as a source:
https://www.filamentgroup.com/lab/select-css.html

Downside is that you have to manage active state, the chevron, disabled state etc. by yourself.

Screenshot 2020-03-31 at 8 34 35
Screenshot 2020-03-31 at 9 16 37

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: Moved array to constants.

PR #47, opened 2020-04-01
OM - 603 | Add subscription view

tuovinensanttu:
Added:
- New view for subscriptions
- New query & mutation for fetching and saving data. (+ Error handling)
- Tests for new component.

For now testing this PR requires local version of tunnistamo and open-city-profile projects. (I'll edit this message after changes are merged to open-city-profile's develop branch)


src/subscriptions/components/subsciptions/Subscriptions.tsx
```diff
@@ -0,0 +1,207 @@
...
+    QueryMySubscriptions
+  >(QUERY_MY_SUBSCRIPTIONS);
+
+  const [updateSubscriptions] = useMutation<
+    UpdateMyProfile,
+    UpdateMyProfileVariables
+  >(UPDATE_PROFILE);
+
+  // Refetch services when language changes, services are translated based on
+  // Accept-Language header which is set in the graphql-client (src/graphql/client).
```
Jokinen, line 72: Interesting that you end up having to do something like this. But the approach you've taken looks very readable an well integrated! Props.

Jokinen (approved):
Wow, you've been busy! Code looked good to me, no gripes. I don't have a local tunnistamo setup yet so I wasn't able to click through the changes.

src/subscriptions/components/subsciptions/Subscriptions.tsx
badrange: Nice solution! 
In kukkuu-ui we were "pragmatic" and solved this by triggering full page reloads when user changes language. This solution will feel much smoother for the end users.

PR #48, opened 2020-04-06
OM - | Added target for footers feedback link.

tuovinensanttu:
Added target for footers feedback link.

### DIFF
--- a/src/i18n/sv.json
+++ b/src/i18n/sv.json
@@ -107,7 +107,8 @@
     "privacy": "Integritetspolicy",
     "accessibility": "Tillgänglighetsdokument",
     "terms": "Användarvillkor",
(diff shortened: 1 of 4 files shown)


src/i18n/sv.json
```diff
@@ -107,7 +107,8 @@
     "privacy": "Integritetspolicy",
     "accessibility": "Tillgänglighetsdokument",
     "terms": "Användarvillkor",
-    "feedback": "Ge feedback"
```
Jokinen, line 110: This is not caps locked while the other values are.

Jokinen (approved):

src/i18n/sv.json
Jokinen: The comment went on the original, when I meant to comment the change.

PR #49, opened 2020-04-08
OM - 521 | Confirm changes before saving

tuovinensanttu:
- If user's profile has service connections on their profile, they are prompted with confirmation modal before saving edited information. 
- Replaced old deleteConfirmationModal with new one.


Jokinen (approved):
Code looks good 👍

PR #50, opened 2020-04-08
OM-686 | Subscription view fixes

tuovinensanttu:
Made following fixes:
- When save is successful, notification is displayed.
- After successful save data is queried again -> saved data is present when navigating between tabs.
- Pressing cancel button resets values to previous.
- Added missing error handling to QueryMySubscriptions.
- Replaced placeholder texts.
- Improved mobile layout.


Jokinen (approved):
Code looked good and happy path worked 👍

PR #51, opened 2020-04-14
OM-673 | Script for updating translations

tuomas777:
Added a script that updates the translation files from a spreadsheet in Google Spreadsheets. It is basically just copy-pasted from [Kukkuu UI](https://github.com/City-of-Helsinki/kukkuu-ui). Also ran the script, so the translation files are now in sync with the spreadsheet and ordered by keys alphabetically.


badrange (approved):
It's really bad to copy scripts from place to place, if someone makes an improvement somewhere, it will not propagate. 

What about creating our first npm package for the script?

tuomas777:
> It's really bad to copy scripts from place to place, if someone makes an improvement somewhere, it will not progress.
> 
> What about creating our first npm package for the script?

Yeah it sure is bad, a npm package is definitely the way to go when someone(tm) has the time to do that 🙂 Then again, IMO we really should look into using an actual service to handle these translations...

badrange:
A proper service is a good idea! Let's bring it up.

PR #53, opened 2020-04-16
OM - 339 |  Configure sentry

tuovinensanttu:
Added sentry for error capturing.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,12 +1,20 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import Modal from 'react-modal';
+import * as Sentry from '@sentry/browser';
 
 import './index.css';
 import BrowserApp from './BrowserApp';
 import * as serviceWorker from './serviceWorker';
 
 Modal.setAppElement('#root');
+
+if (process.env.NODE_ENV !== 'development') {
+  Sentry.init({
+    dsn: process.env.REACT_APP_SENTRY_DSN,
+    environment: process.env.NODE_ENV,
+  });
+}
 
 ReactDOM.render(<BrowserApp />, document.getElementById('root'));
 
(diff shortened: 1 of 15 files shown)


src/index.tsx
```diff
@@ -1,13 +1,21 @@
...
 import Modal from 'react-modal';
+import * as Sentry from '@sentry/browser';
 
 import './index.css';
 import BrowserApp from './BrowserApp';
 import * as serviceWorker from './serviceWorker';
 
 Modal.setAppElement('#root');
 
+if (process.env.NODE_ENV !== 'development') {
```
Jokinen: It's application dependant which values for `NODE_ENV` are possible. I didn't go through the code to find out if this comment makes much of a difference with this project. The possible values for `NODE_ENV` aren't always limited to `development` and `production`.  For instance, sometimes testing frameworks or apps use `NODE_ENV=test`. You wouldn't ideally enable Sentry in this case, though it's unlikely that it'd cause issues either.

Whitelisting instead of blacklisting could be more explicit. Here's one pattern you can use to whitelist multiple `NODE_ENV`s.
```js
const ENVS_WITH_SENTRY = ['production', 'staging'];

ENVS_WITH_SENTRY.includes(process.env.NODE_ENV)
```

Jokinen (approved):
Code looked good to me 👍

src/index.tsx
tuovinensanttu: I think this is a good point. I'm not sure if the tests have ever triggered sentry, but I made the change so I remember this in the future as well.

badrange: Note that NODE_ENV will always be production also on our staging servers, since the variable is set by create-react-app during build, and we have production builds on staging.

badrange: Which means that it would be better to set the check to === production, I'll make that change in kukkuu, good point about test-env!

tuovinensanttu: In that case I'll make the change as well.

PR #54, opened 2020-04-21
OM-690 | Profile language is displayed on "My information" page

tuovinensanttu:
- Added profile language to "My information" page.
- Added value field to select component. This resolves issue where profile language value wasn't updated on edit form.


Jokinen (approved):
Changes look good to me 👍

badrange (approved):
No snapshot test for this component, I see. At some point, on a rainy day, we could decide which components that are missing tests are the most critical and start adding some..

PR #55, opened 2020-04-22
OM-696 | Download data button works only once

tuovinensanttu:
By adding `fetchPolicy: 'network-only'` the onCompleted method is run every time query is called.


Jokinen (approved):
Code makes sense. I also tested the happy path and it worked 👍

PR #56, opened 2020-04-24
OM-738 | Remove terms page

tuovinensanttu:
Removed everything related to terms-of-service page. Registration page "Confirm text" will be changed / replaced at later date.


badrange (approved):
I love it when we delete lots of stuff!

badrange (approved):

PR #58, opened 2020-04-28
OM-739 | Block site from search engines

tuovinensanttu:
Blocked site from search engines & cleaned up index.html a little bit.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,12 +1,12 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="fi">
   <head>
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -8,34 +8,11 @@
       name="description"
```
badrange, line 8: The page title is now Profiili, but meta description and html lang are in English. We're going to get a complaint from the accessibility review html lang is not in sync with the meta tags and other content

badrange (approved):
One change was missing. 

Me like the cleanup!

PR #59, opened 2020-05-04
OM-740 | Add analytics

tuovinensanttu:
Added analytics to project.


src/common/pageLayout/PageLayout.tsx
```diff
@@ -8,9 +10,24 @@ import styles from './PageLayout.module.css';
...
+  const { title = 'appName' } = props;
+
+  const pageTitle =
+    props.title !== 'appName' ? `${t(title)} - ${t('appName')}` : t('appName');
+
+  useEffect(() => {
+    trackPageView({
+      documentTitle: pageTitle,
+      href: window.location.href,
+    });
```
Jokinen, line 28: This `useEffect` has the same issue I mentioned in the previous PR. I'm not entirely sure, but I think that you had trouble with the linter then so I am highlighting again in case that's the case here as well.

Jokinen (approved):
Looks good to me 👍

src/common/pageLayout/PageLayout.tsx
tuovinensanttu: Nice catch. Seems i'm relying a little too much on linting 😅

PR #60, opened 2020-05-05
OM-632 |  Change terms text

tuovinensanttu:
Task was to replace "I have read the terms..." text with a new one, but because I used the update-translations script, couple other texts got updated as well.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -92,7 +92,7 @@
     "phone": "Phone",
     "postalCode": "Postal code",
     "submit": "Save",
-    "terms": "I have red <0>file description</0> and  city's <1>data protection principles</1>"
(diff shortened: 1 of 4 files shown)


src/i18n/en.json
```diff
@@ -93,7 +92,7 @@
     "phone": "Phone",
     "postalCode": "Postal code",
     "submit": "Save",
-    "terms": "I agree to the <0>Terms of Service</0>, and that I may be contacted at the address I provide"
+    "terms": "I have red <0>file description</0> and  city's <1>data protection principles</1>"
```
badrange: red should be read :-P

badrange (approved):
I think it is ok to merge translation changes without waiting for pull request approval, at the same time I noticed a speling error..

PR #61, opened 2020-05-07
Use fetch-translations script from helsinki-utils

Jokinen:
Replaces copy pasted script with imported one from `helsinki-utils`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,14 +71,11 @@
     ]
   },
   "devDependencies": {
-    "@types/node-fetch": "^2.5.6",
     "apollo": "^2.21.1",
     "eslint-config-prettier": "^6.4.0",
     "eslint-plugin-prettier": "^3.1.1",
     "helsinki-utils": "City-of-Helsinki/helsinki-utils-js#0.1.0",
-    "i18next-json-csv-converter": "^0.2.0",
     "jest-fetch-mock": "^2.1.2",
-    "node-fetch": "^2.6.0",
     "prettier": "^1.18.2",
     "ts-node": "^8.8.2"
   }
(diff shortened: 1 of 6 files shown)


tuovinensanttu (approved):
Looks good 👍 I think it's a nice idea that you used another approach here. Now there is two working examples how to use this new utility.

package.json
```diff
@@ -53,7 +53,7 @@
     "ci": "CI=true yarn test --coverage",
     "lint": "eslint --ext js,ts,tsx src",
     "codegen": "apollo client:codegen ./src/graphql/generatedTypes.ts --outputFlat --includes=./src/**/*.graphql --target=typescript --endpoint=https://profiili-api.test.kuva.hel.ninja/graphql/ --useReadOnlyTypes --addTypename",
-    "update-translations": "ts-node  -P ./scripts/tsconfig.json -r dotenv/config scripts/update-translations.ts"
+    "update-translations": "ts-node -P ./scripts/tsconfig.json -r dotenv/config --files scripts/update-translations.ts"
```
badrange, line 56: Didn't know about this way to read config, nice!

scripts/helsinki-utils.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'helsinki-utils/scripts/fetch-translations' {
```
badrange, line 1: Did you consider adding the types in a PR towards helsinki-utils?

badrange (approved):
This solution is fine! 
If you remove the local script that calls the one from the library you might get away from having ts-node as a dependency of this project (unless there is a need from somewhere else..)

scripts/helsinki-utils.d.ts
Jokinen: Yup 👍I wrote down an issue to track it: https://github.com/City-of-Helsinki/helsinki-utils-js/issues/5

Jokinen:
By doing something magical, I've managed to install the package so that it has been recorded within the lock file, but not within package.json itself. I'll fix this before merge.

Jokinen:
@badrange The original script pulled some configs from the ENV. I wanted to avoid removing that functionality, because I was not sure how it was used. I felt that this was the quickest way to retain that functionality.

But you are definitely correct. If the configuration is not used, simplifying the script would remove complexity in configuring, dependencies and code.

Jokinen:
I had also forgotten to remove those dependencies that are now depended on by the `helsinki-utils` repo. `node-fetch` (+types) and `i18n-csv-to-json-converter`. I've removed these.

PR #62, opened 2020-05-11
OM-716 |  Add county code

tuovinensanttu:
- Added country to profile detail view and edit profile form
- Adding new field to mutations caused one of the Subscription types to throw TS error, fixed this as well.
- Updated translations.

### DIFF
--- a/src/profile/helpers/getAddress.ts
+++ b/src/profile/helpers/getAddress.ts
@@ -1,13 +1,6 @@
 import countries from 'i18n-iso-countries';
-import fi from 'i18n-iso-countries/langs/fi.json';
-import en from 'i18n-iso-countries/langs/en.json';
-import sv from 'i18n-iso-countries/langs/sv.json';
 
 import { MyProfileQuery } from '../../graphql/generatedTypes';
-
-countries.registerLocale(fi);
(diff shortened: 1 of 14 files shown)


src/profile/helpers/getAddress.ts
```diff
@@ -1,9 +1,20 @@
+import countries from 'i18n-iso-countries';
+import fi from 'i18n-iso-countries/langs/fi.json';
+import en from 'i18n-iso-countries/langs/en.json';
+import sv from 'i18n-iso-countries/langs/sv.json';
+
 import { MyProfileQuery } from '../../graphql/generatedTypes';
 
-export default function getAddress(data: MyProfileQuery) {
+countries.registerLocale(fi);
```
Jokinen: You are configuring this module in the same way in two places 🤔Maybe there's a petter which would allow you to only do it once?

Jokinen (approved):
Code side looked good 👍

src/profile/helpers/getAddress.ts
tuovinensanttu: Good point. I moved registerLocale stuff to App.tsx, seems to be working.

badrange (approved):

PR #67, opened 2020-05-19
OM-807 | Use checkbox from HDS, make explanations flush and fix broken errors

Jokinen:
- Removes this project specific checkbox and uses the one from HDS instead
- Makes explanation flush for profile edit form--previously the explanation and the form were not flush on the y-axis
- Fixes broken error messages with inputs: HDS deprecated `invalidText` prop for `TextInput`


src/profile/helpers/formik.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return isSubmitted && isError;
+}
+
+const defaultErrorRender = (error: string) => error;
+
+export function getError<FormValues>(
+  formikProps: FormikProps<FormValues>,
+  fieldName: keyof FormValues,
+  render: (error: string) => React.ReactNode = defaultErrorRender
```
tuovinensanttu: You are using `React.ReactNode` here but your import says `{ ReactNode } from 'react'`

src/common/explanation/Explanation.tsx
```diff
@@ -7,13 +7,23 @@ type Props = {
   main: string;
   small?: string;
   className?: string;
+  variant?: 'margin' | 'flush';
```
badrange, line 18: It isn't obvious to an outsider what the meaning of flush is, so some kind of comment could be helpful!

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -87,8 +87,11 @@ function DeleteProfile(props: Props) {
 
         <Checkbox
           onChange={handleDeleteInstructions}
+          id="deleteInstructions"
           name="deleteInstructions"
-          label={
+          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+          // @ts-ignore
```
badrange, line 94: Do we have an issue or a plan for how we can remove the need for these ts-ignores?

badrange (approved):

tuovinensanttu:
Screenshot 2020-05-19 at 10 50 27

Seems like something is breaking the layout below 470px (width)

src/profile/components/deleteProfile/DeleteProfile.tsx
Jokinen: I've opened an issue about this in HDS.

src/common/explanation/Explanation.tsx
Jokinen: Added comment.

src/profile/helpers/formik.ts
Jokinen: Fixed 👍

Jokinen:
@tuovinensanttu This is how it looks for me

It looks like your browser may not support text ellipsis (or my browser, chrome, applies it automatically). But there are other inconsistencies as well--so I am not sure.

Could this relate to something along these lines: with the previous version of HDS, HDS used smaller font sizes for smaller screens and it no longer does that?

Jokinen:
We chatted on Slack and I was able to find the root of the problem 👍

PR #68, opened 2020-05-19
OM-821 | Change translation sheet ID to point to the XO sheet

klempine:
XO took control of the translation sheet for Helsinki profile UI, so I
changed the config to point to the new sheet ID.


Jokinen (approved):
Looks good 👍

PR #70, opened 2020-05-20
OM-786 | Add accessibility docs

tuovinensanttu:
Added accessibility statement.


src/accessibilityStatement/AccessibilityStatement.module.css
```diff
@@ -0,0 +1,53 @@
+.pageWrapper {
+    min-height: 100vh;
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+}
+
+.container {
+    margin: 20px 10px 0;
```
Jokinen: I am thinking if we should not start using spacing variables from HDS now that they are relatively stable?

src/accessibilityStatement/AccessibilityStatement.module.css
```diff
@@ -0,0 +1,53 @@
...
+.innerWrapper h1 {
+    font-size: var(--fontsize-h-3);
+}
+
+.innerWrapper h2 {
+    margin-top: 25px;
+    font-size: var(--fontsize-h-4);
+}
+
+@media (min-width: 800px) {
```
Jokinen: There are different breakpoints in the app, but this is the only one of `800px`. The closes value currently in use would be `768px`. I don't know if it makes a huge difference for this case.

src/accessibilityStatement/AccessibilityStatement.tsx
```diff
@@ -0,0 +1,39 @@
...
+import Header from '../common/header/Header';
+import Footer from '../common/footer/Footer';
+import AccessibilityStatementFi from './AccessibilityStatementFi';
+import AccessibilityStatementSv from './AccessibilityStatementSv';
+import AccessibilityStatementEn from './AccessibilityStatementEn';
+
+function AccessibilityStatement() {
+  const { i18n } = useTranslation();
+  const selectStatement = () => {
+    const lang = i18n.languages[0];
```
Jokinen: Might this language value also sometimes contain the locale information? If I remember correctly you configured locales with both youth-membership and profile.

src/accessibilityStatement/AccessibilityStatementFi.tsx
```diff
@@ -0,0 +1,106 @@
...
+        palautelomakkeella{' '}
+        <a href="https://www.hel.fi/palaute">www.hel.fi/palaute</a>.
+        Tiedusteluun pyritään vastaamaan kohtuullisessa ajassa.
+      </p>
+
+      <h2>Palaute ja yhteystiedot</h2>
+      <p>
+        Kaupunginkanslia, Helsinki
+        <br />
+        +358 9 310 1641
```
Jokinen: A `tel:` link is possible here. But maybe it has been omitted for a specific reason?

Jokinen (approved):
Looks good to me 👍 A few comments.

src/accessibilityStatement/AccessibilityStatement.module.css
tuovinensanttu: Sure, I'll change these to use hds spacing.

src/accessibilityStatement/AccessibilityStatement.module.css
tuovinensanttu: I'll change this to `768` to keep things consistent.

src/accessibilityStatement/AccessibilityStatement.tsx
tuovinensanttu: You are totally right, nice catch 👍

src/accessibilityStatement/AccessibilityStatementFi.tsx
tuovinensanttu: Added `tel:` link.

badrange (approved):
👍

PR #72, opened 2020-05-20
OM-806 | Mobile navigation

tuovinensanttu:
Hide logged in user links if user is not logged in.

### DIFF
--- a/src/common/fullscreenNavigation/FullscreenNavigation.tsx
+++ b/src/common/fullscreenNavigation/FullscreenNavigation.tsx
@@ -42,17 +42,21 @@ function FullscreenNavigation(props: Props) {
           />
         </div>
         <div className={styles.navItems}>
-          <NavLink to="/" className={styles.navLink} onClick={closeNav}>
-            {t('nav.information')}
-          </NavLink>
-          <NavLink
-            to="/connected-services"
-            className={styles.navLink}
-            onClick={closeNav}
-          >
-            {t('nav.services')}
-          </NavLink>
           {isAuthenticated && (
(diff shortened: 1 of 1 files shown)


src/common/fullscreenNavigation/FullscreenNavigation.tsx
```diff
@@ -42,17 +42,21 @@ function FullscreenNavigation(props: Props) {
...
-          </NavLink>
-          <NavLink
-            to="/connected-services"
-            className={styles.navLink}
-            onClick={closeNav}
-          >
-            {t('nav.services')}
-          </NavLink>
           {isAuthenticated && (
+            <React.Fragment>
```
Jokinen, line 46: Shouldn't our version of babel support `<>`? Do you simply prefer this notation?

Jokinen (commented):
Looks to be working 👍

src/common/fullscreenNavigation/FullscreenNavigation.tsx
tuovinensanttu: I just prefer it 😄, using `Fragment` just makes it feel 'cleaner', if that makes any sense...

Jokinen (approved):
Sorry, I made a comment when I meant to approve :D

src/common/fullscreenNavigation/FullscreenNavigation.tsx
badrange: I prefer <> - wonder if we're going to start a vi vs emacs war 😄

PR #73, opened 2020-05-20
OM-807 | Fix last style details on profile view page

Jokinen:
Screenshot 2020-05-20 at 10 59 50

* Makes buttons full width on mobile
* Adds whitespace to expanding boxes as per the abstract files (discussed with Anniina)
* Restyles profile information edit button
* Moves language as the last field (additional comment from Anniina)
* Renders address on a single row (additional comment from Anniina)

---

The styles between the information and edit views were out of sync. I fixed this by creating a new reusable component that applies the styles and some layout rules. This guards against styles getting out of sync again.


badrange (approved):
👍

tuovinensanttu:
Screenshot 2020-05-20 at 14 23 44

Seems like we missed some errors in console. These errors are caused by changes made in #67.

tuovinensanttu:
Screenshot 2020-05-20 at 14 29 07

If I understood correctly this was the last set of changes for OM-807. Maybe add more margin between Checkbox and button?

src/profile/helpers/getAddress.ts
```diff
@@ -5,9 +5,11 @@ import { MyProfileQuery } from '../../graphql/generatedTypes';
...
   if (data.myProfile?.primaryAddress) {
     const address = data.myProfile.primaryAddress;
-    return `${address.address}, ${address.city} ${
-      address.postalCode
-    } \n ${countries.getName(address.countryCode || 'FI', lang)}`;
+    const country = countries.getName(address.countryCode || 'FI', lang);
+
+    return [address.address, address.city, address.postalCode, country].join(
+      ', '
+    );
```
tuovinensanttu, line 12: You don't have to fill every field in address. For example filling only postalCode will result in `, , 12345, Finland`.

Jokinen: Oh ok, the previous implementation also had this issue with missing address/city so I ignored it. We can just filter `undefined` values out in which case the result does not have extra commas. It may result in weird addresses like "Finland" though. Would that be an issue 🤔

Jokinen:
GitHub lagged so I only saw Bent's comment! I reverted the changes, will fix Santtu's points and post a new PR.

src/profile/helpers/getAddress.ts
tuovinensanttu: True, it was an issue before. I just thought that if we are making improvements to UI, it would be nice to fix this as well :)

PR #74, opened 2020-05-20
OM-808 |  Mobile download data

tuovinensanttu:
Downloading my data should now working on iOS as well. I tested functionality with simulator and it seems to be working.

Testing:
I added button to login page, which can be used for testing with real device. This button will be removed before merging to develop.
EDIT: This button is now removed.


Jokinen (approved):
Looks good to me 👍

PR #76, opened 2020-05-22
OM-807 | Fix last style details on profile view page

Jokinen:
I've addressed the comment in https://github.com/City-of-Helsinki/open-city-profile-ui/pull/73 in this PR.

---

Screenshot 2020-05-20 at 10 59 50

* Makes buttons full width on mobile
* Adds whitespace to expanding boxes as per the abstract files (discussed with Anniina)
* Restyles profile information edit button
* Moves language as the last field (additional comment from Anniina)
* Renders address on a single row (additional comment from Anniina)

---

The styles between the information and edit views were out of sync. I fixed this by creating a new reusable component that applies the styles and some layout rules. This guards against styles getting out of sync again.


src/common/explanation/Explanation.tsx
```diff
@@ -30,7 +38,14 @@ function Explanation({ className, main, small, variant = 'margin' }: Props) {
         className
       )}
     >
-      <h1 className={styles.main}>{main}</h1>
+      <h1
```
badrange, line 41: We really need to get the header sizes upstream in HDS sorted!!

badrange: Maybe we can add some kind of TODO here so it is easy to find the spots that needs fixing once it happens..

badrange (approved):

src/common/explanation/Explanation.tsx
badrange: I created an issue in HDS: https://github.com/City-of-Helsinki/helsinki-design-system/issues/143

badrange: And maybe this is not the best example of this issue.

Jokinen: Yeah, I think that we may still have some things to work out within this app 🤔 I'm not sure if all the H1 element we have actually need to be H1 elements. Then some of the H1 element we have are smaller than some other visual headings we have. But yeah, I wasn't able to get my head around it enough that I would have liked to make the change here.

PR #79, opened 2020-05-25
OM-831 | Fix sentry config

tuovinensanttu:
Fix sentry environment always pointing to production.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,7 +12,8 @@ Modal.setAppElement('#root');
 const ENVS_WITH_SENTRY = ['staging', 'production'];
 
 if (
-  ENVS_WITH_SENTRY.includes(process.env.REACT_APP_ENVIRONMENT || 'development')
+  process.env.REACT_APP_ENVIRONMENT &&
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -11,10 +11,12 @@ Modal.setAppElement('#root');
 
 const ENVS_WITH_SENTRY = ['staging', 'production'];
 
-if (ENVS_WITH_SENTRY.includes(process.env.NODE_ENV)) {
+if (
+  ENVS_WITH_SENTRY.includes(process.env.REACT_APP_ENVIRONMENT || 'development')
```
Jokinen: Not a huge deal, but this may create inconsistencies in the long run. In other words,  when `REACT_APP_ENVIRONMENT` environment is consumed, it's `undefined` (etc.) in some place and `development` here, because you've set a default value. Having a config file that does the default value handling would allow you to enforce consistency between the users of this env.

Jokinen (approved):
Looks good to me 👍

src/index.tsx
tuovinensanttu: Good point, I made a little tweak to this.

PR #80, opened 2020-05-25
OM-783 | Update the text content

klempine:
Updated the latest text content from the translation Excel.


Jokinen (approved):
Looks good 👍I took the liberty to add a commit that removes a link definition that is no longer used after these changes.

klempine:
> 
> 
> Looks good 👍I took the liberty to add a commit that removes a link definition that is no longer used after these changes.

Thank you sir!

PR #82, opened 2020-05-27
OM-335 | Fix silent renew

Jokinen:
Note that in order to have this work, the `silent_redirect_uri` needed to be added into the `redirect_uris` config at Tunnistamo's end (the correct instance of it).

Silent renewal should now work locally. It won't work in the PR build nor will it work in staging (or production yet). I'll ask for the staging configs to be set once I know that these changes are correct so I don't end up bothering Ville too much 😄Prod configs will be deferred for now.

### DIFF
--- a/public/silent_renew.html
+++ b/public/silent_renew.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+</head>
+<body>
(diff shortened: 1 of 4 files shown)


tuovinensanttu (approved):
Gave it a little test and it seems to be working. Nice 👍

badrange (approved):
Looks good to me too! 

That hard coded oidc-client version in silent_renew.html is going to bite us hard at some point.

PR #83, opened 2020-05-27
Release/1.0.0 rc.2

klempine:
Merged the current changes in develop into master to do a release into the production environment.


tuovinensanttu (approved):
I think version in `package.json` should be updated to new version. Otherwise LGTM 👍

klempine:
Thanks for spotting that one, I changed the value now.

PR #84, opened 2020-05-28
OM-720 |  Multiple emails

tuovinensanttu:
- Added support for adding multiple emails.
- Created helper for creating mutation variables object
- Cleaned up `EditProfileForm.module.css` for some reason there was same media query three times.


src/profile/components/editProfileForm/EditProfileForm.module.css
```diff
@@ -41,21 +19,64 @@
...
 }
 
+.removeButton {
+    margin: 5px 0 0 3px;
+    color: var(--color-bus)
+}
+
 .button {
-    margin-top: 20px;
+    margin-top: 10px;
```
Jokinen: Up to you, but here's one chance to sync these values up with HDS.

src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -211,11 +218,58 @@ function EditProfileForm(props: Props) {
...
+                      (email: Email | PrimaryEmail, index: number) => (
+                        <div key={index} className={styles.formField}>
+                          <Field
+                            as={TextInput}
+                            name={`emails.${index}.email`}
+                            id={`emails.${index}.email`}
+                            labelText={t('profileForm.email')}
+                          />
+                          {!email.id && (
+                            <button
```
Jokinen, line 258: Looking at the code, I get the sense that this button may be styled in a weird way on some platforms, for instance mobile. As far as I can tell, you are using the default button element with one class that just adjusts margin and colour 🤔If buttons are reset in a normalize.css etc, then I can see how it could work.

src/profile/helpers/getEmailsFromNode.ts
```diff
@@ -0,0 +1,26 @@
...
+            email: edge.node.email,
+            id: edge.node.id,
+            primary: edge.node.primary,
+            emailType: edge.node.emailType,
+          } as Email;
+        }
+        return false;
+      })
+      .filter((email): email is Email => Boolean(email))
+      .filter((email: Email) => !email.primary);
```
Jokinen: I think that this is sort of hard to read 🤔Here's a quick suggestion on how you can make it a bit simpler.
```js
edge
  .filter(edge => edge.node && edge.node.primary)
  .map(edge => ({
    email: edge.node.email,
    id: edge.node.id,
    primary: edge.node.primary,
    emailType: edge.node.emailType,
  }))
```

You could do something like
```js
const edges = data?.myProfile?.emails?.edges || []; // don't know the correct TS pattern
```
to avoid multiple returns as well.

src/profile/helpers/updateMutationVariables.ts
```diff
@@ -0,0 +1,119 @@
...
+        },
+      ],
+    };
+  }
+
+  return {
+    addAddresses: [
+      formValues.address || formValues.postalCode || formValues.city
+        ? {
+            address: formValues.address,
```
Jokinen, line 33: Lot's of these rows are the same as compared to the first return. You've used the same pattern a few times in this file. Imo it makes maintenance a bit harder because you have potentially have more places you have to edit.

Jokinen (approved):
Looks good to me 👍I left some non vital code style comments.

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: It seems to working on mobile resolution, nothing is exploding.

src/profile/helpers/getEmailsFromNode.ts
tuovinensanttu: Thanks for suggestion! I was struggling to find a solution that would preserve required types.

src/profile/components/editProfileForm/EditProfileForm.tsx
Jokinen: I think you'd have to use a mobile browser. For instance iOS Safari tends to force some sort of background and borders 🤔

tuovinensanttu: I'll give that a go before merging.

src/profile/helpers/getEmailsFromNode.ts
Jokinen: Oh, ok. I imagined there was something like that going on! I have no idea if types are retained with that solution.

src/profile/components/editProfileForm/EditProfileForm.tsx
Jokinen: But maybe if there's nothing on Chrome the chance is quite small.

src/profile/helpers/updateMutationVariables.ts
tuovinensanttu: Thats true. I tried to throw all of these fields to variable and use spread operator, that however results typing breaking (it can't find ID for some reason 😅 ). These methods are going to be replaced anyway when support for multiple addresses and phones are added.

src/profile/helpers/getEmailsFromNode.ts
tuovinensanttu: Seems to be working just fine. I think my mistake was to add `if` check inside map, that resulted typing to be `(Email | undefined)[]`.

PR #87, opened 2020-06-02
OM-789 |  Remove emails

tuovinensanttu:
Removing emails is now possible. 

Made a little tweak based on comments:
- Additional emails that are saved will be shown in readOnly mode
- Added PlusCircle icon to "Add additional emails" button.


src/common/svg/PlusIcon.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
Jokinen: Not a huge deal, but this SVG has unnecessary meta info in it. If you run it through a svg minifier, these will be removed.

Jokinen (approved):
Looks good to me 👍

PR #89, opened 2020-06-02
OM-832 | Fix message being cut off

tuovinensanttu:
- Updated HDS
- Added more margin between notification title and message => text is no longer hidden below close button.


Jokinen (approved):
Should this be reported to HDS?

PR #91, opened 2020-06-04
OM-795 | User can change primary email

tuovinensanttu:
- Changing primary email is now possible
- Updated HDS (again...) to get access to new `IconCirclePlus`. Update caused a little side effect, I needed to fix few other icons here and there.
- Email validation is done with Validator. [Test cases](https://github.com/validatorjs/validator.js/blob/10070979e36d3e9a977e4abccfd0e2be73070cb9/test/validators.js#L60) seem to overlap pretty nicely with django.
- `getEmails` should be pretty well tested.
- Had some user errors with `getEmailsFromNode` function, so I decided to add tests to that as well.


Jokinen (approved):
Looks good to me 👍

PR #93, opened 2020-06-04
Release/1.0.0 rc.3

klempine:
Merged the current changes from develop into master to do a deployment into production env. The most important thing about the changes is that the UI should now be pointing to api.hel.fi/profiili which is the official URL for the API. We'll try this in production to be sure that it works.


tuovinensanttu (approved):
LGTM 👌

PR #94, opened 2020-06-04
OM-841 | User session is handled across multiple tabs.

tuovinensanttu:
Logging in or out will affect user session across multiple tabs. Same changes as in [Jässäri](https://github.com/City-of-Helsinki/youth-membership-ui/pull/99)


Jokinen (approved):
Looks good to me 👍

PR #96, opened 2020-06-05
OM-631 | Add file description link to footer

tuovinensanttu:
- Footer link to file description now points to correct address
- Fixed file description and data protection links. They were wrong way around 🤦‍♂️


Jokinen (approved):
Looks good to me 👍

PR #99, opened 2020-06-11
OM-722 | Add multiple addressess

tuovinensanttu:
- Multiple address support (add, edit, delete, view)
- Wrote same kind of tests for `getAddresses` as `getEmails`
- Country select is still using projects custom select instead of HDS one. Seems like the HDS's Dropdown doesn't play along with formik out of the box. I'll make the change in separate PR.

### DIFF
--- a/src/common/test/myProfileQueryData.ts
+++ b/src/common/test/myProfileQueryData.ts
@@ -1,4 +1,5 @@
 import {
+  AddressType,
   EmailType,
   Language,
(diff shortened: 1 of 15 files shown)


Jokinen (commented):
Code looks good. I'll try out the UI quickly as well.

Jokinen (approved):
Looks to be working nicely 👍

PR #100, opened 2020-06-12
OM-721 | Multiple phones

tuovinensanttu:
- Added multiple phone support: user can add, view, edit, remove and change primary phones. 
- Made changes to `updateMutationVariables` helper function. It's now as generic as possible.
- Updated / added tests.

### DIFF
--- a/src/profile/helpers/__tests__/updateMutationVariables.test.ts
+++ b/src/profile/helpers/__tests__/updateMutationVariables.test.ts
@@ -1,21 +1,23 @@
 import {
   AddressType,
   EmailType,
-  MyProfileQuery,
+  Language,
+  UpdateMyProfileVariables,
   MyProfileQuery_myProfile_addresses_edges_node as Address,
   MyProfileQuery_myProfile_emails_edges_node as Email,
+  MyProfileQuery_myProfile_phones_edges_node as Phone,
+  PhoneType,
 } from '../../../graphql/generatedTypes';
-import { getEmail, getAddress } from '../updateMutationVariables';
+import { FormValues } from '../../components/editProfileForm/EditProfileForm';
+import { updateMutationVariables } from '../updateMutationVariables';
 import { myProfile } from '../../../common/test/myProfileQueryData';
-
-// TODO add tests for getAddress & getEmail after support for multiple entities is added
 
 const emails: Email[] = [
   {
     email: 'ensimmainen@testi.fi',
-    primary: true,
-    id: '123',
     emailType: EmailType.OTHER,
+    id: '123',
+    primary: true,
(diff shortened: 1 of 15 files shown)


src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -66,113 +69,176 @@ const addresses: Address[] = [
...
-  });
-
-  test('removeEmails exists', () => {
-    const emailObj = getEmail([], myProfile);
-    expect(emailObj.removeEmails).toEqual(['123', '234']);
-  });
+const phones: Phone[] = [
+  {
+    id: '123',
+    phone: '0501234567',
```
badrange, line 75: Do we have any tests with foreign phone numbers or addresses?

src/profile/helpers/updateMutationVariables.ts
```diff
@@ -28,139 +36,146 @@ type AddressInputs = {
...
-      );
+const getPrimaryValue = (primary: Primary, profile?: MyProfileQuery) => {
+  const primaryValue = profile?.myProfile && profile.myProfile[primary];
+  return primaryValue || { id: '' };
+};
 
-      return address.id && !isEqual(address, profileAddress);
-    })
-    .map(address => {
+const getEmptyObject = (primary: Primary) => {
```
badrange, line 53: It's not obvious what this is for, maybe a comment explaining the purpose could be helpful!

badrange (approved):
Looks solid! 

Could be a good idea to add a foreign phone number for the tests and some code comments.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
tuovinensanttu: There isn't currently any validation for phone numbers and the only address based validation is that postalCode is at the most 5 characters. So at the moment every phone number will pass. But if we end up adding stricter validation, testing foreign values is a good idea!

src/profile/helpers/updateMutationVariables.ts
tuovinensanttu: I'll add a comment. My imagination was running on fumes when I was naming these 😄

Jokinen (approved):
Bernt beat me to it! Looking good as far as I am concerned.

One thing I am thinking about is the UX in the view for additional fields. The controls are at the very bottom, but the new fields are always generated in a grouped order where emails are first, then addresses and so on. So if a user has an additional address and they add an email, it may be enough distance away from the button that it's not obvious.

I imagine Anniina will give more details on these at some point so I don't think it's something we should address now. Just barfing my thoughts :D

PR #102, opened 2020-06-15
OM-722 | Adding address causes backend error

tuovinensanttu:
Added typenames to empty primary and phone objects.  This way `getObjectFields` function will never return default value and cause empty object appearing in `addValues` array.

Also added `primary` field to primaryPhone. This fixed issue where adding a primaryPhone in edit form would cause it to be added as secondary.

### DIFF
--- a/src/profile/components/editProfileForm/EditProfileForm.tsx
+++ b/src/profile/components/editProfileForm/EditProfileForm.tsx
@@ -150,12 +150,16 @@ function EditProfileForm(props: Props) {
           address: props.profile.primaryAddress.address || '',
           postalCode: props.profile.primaryAddress.postalCode || '',
           city: props.profile.primaryAddress.city || '',
+          // User can not add address when registering. Set primary field to true by default
+          // so it gets added to correct place
           primary: props.profile.primaryAddress.primary || true,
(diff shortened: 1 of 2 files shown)


src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -152,10 +152,13 @@ function EditProfileForm(props: Props) {
           city: props.profile.primaryAddress.city || '',
           primary: props.profile.primaryAddress.primary || true,
           countryCode: props.profile.primaryAddress.countryCode || 'FI',
+          __typename: props.profile.primaryAddress.__typename || 'AddressNode',
         },
         primaryPhone: {
           ...props.profile.primaryPhone,
+          primary: props.profile.primaryPhone.primary || true,
```
badrange, line 163: A comment in the code explaining why this defaults to true could be helpful.

badrange (approved):

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: Added comments to both primary address and phone.

PR #103, opened 2020-06-15
OM-722 | Add validation to additional address fields

tuovinensanttu:
- Errors occurring while adding additional phones / addresses are now displayer correcly.
- Fixed issue where changing phone number to primary caused errors.

### DIFF
--- a/src/profile/components/editProfileForm/EditProfileForm.tsx
+++ b/src/profile/components/editProfileForm/EditProfileForm.tsx
@@ -368,6 +368,21 @@ function EditProfileForm(props: Props) {
                             name={`phones.${index}.phone`}
                             id={`phones.${index}.phone`}
                             labelText={t('profileForm.phone')}
+                            type="tel"
+                            minLength="6"
+                            maxLength="255"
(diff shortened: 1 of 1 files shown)


src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -406,25 +422,61 @@ function EditProfileForm(props: Props) {
...
+                              `addresses.${index}.address`,
+                              {
+                                max: 128,
+                              }
+                            )}
                           />
 
                           <Field
                             className={styles.formField}
+                            maxLength="5"
```
badrange, line 444: Is it so that this covers all the postal codes in all the world's countries?

badrange (approved):

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: I think the improved postal code validation will be added later. Idea for now is that it just does the same thing as primary addresses postal code field.

PR #105, opened 2020-06-17
OM-856 | Fix save button stays disabled

tuovinensanttu:
If backend error occurs while saving formik's `isSubmitting` will stay `true`. Submit buttons disabled state is now only controlled by mutation hooks `loading`.

### DIFF
--- a/src/profile/components/createProfileForm/CreateProfileForm.tsx
+++ b/src/profile/components/createProfileForm/CreateProfileForm.tsx
@@ -164,11 +164,7 @@ function CreateProfileForm(props: Props) {
           <div>
             <Button
               type="submit"
-              disabled={Boolean(
-                formikProps.isSubmitting ||
-                  formikProps.errors.terms ||
-                  props.isSubmitting
-              )}
+              disabled={Boolean(formikProps.errors.terms || props.isSubmitting)}
             >
               {t('profileForm.submit')}
             </Button>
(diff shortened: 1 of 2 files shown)


Jokinen (approved):

Jokinen:
GH is doing something weird :D 

But I had a comment about this:

This leaves Formik's state in a broken state. I think you should fix it, because devs familiar with Formik expect the `isSubmitting` prop to work.

If `onValues` ~is~ retusn a promise, Formik will automatically set `isSubmitting` once it resolves. Formik also provides the "formik bag" to `onSubmit` which containes `setSubmitting` that can be used to control the `isSubmitting` state.

tuovinensanttu:
> GH is doing something weird :D
> 
> But I had a comment about this:
> 
> This leaves Formik's state in a broken state. I think you should fix it, because devs familiar with Formik expect the `isSubmitting` prop to work.
> 
> If `onValues` is a promise, Formik will automatically set `isSubmitting` once it resolves. Formik also provides the "formik bag" to `onSubmit` which containes `setSubmitting` that can be used to control the `isSubmitting` state.

Hmm, good point. I'll take another look at this.

badrange (approved):
👍

tuovinensanttu:
Changed formik's `onSubmit` to be async function. This will set `isSubmitting` to false automatically after `onSubmit` resolves.

Jokinen (approved):
Tested the UI and the bug has been fixed 👍

PR #108, opened 2020-06-24
OM-871 | Provide OAuth/Oidc authorize code as part of the GDPR API related calls

Jokinen:
This PR introduces a mechanism which allows the UI to request authorization codes from Tunnistamo, which profiili backend can then use to make authorized requests to other services.

Let's take downloading profile data as an example.

1) User clicks the download button
2) User is redirected to Tunnistamo which, if necessary, renders an UI the user can use to allow a set of permissions (currently these are empty)
3) User is redirected back to profiili UI and the download action is completed

In essence, we need to request the authorization code in the UI, because the user flow may contain a step that requires user input. Once this code is generated, it must be provided within the download profile query and delete profile mutation when these requests are sent to the profile backend. The backend can then use this code to make requests to all the other services for data or deletion.

## Technical jargon

This flow introduces one difficult step--the exit and re-entry into the profile UI application. This makes the download and deletion code flows more difficult to a degree. In comparison, these actions are currently completed with callbacks and promises--which make use of the fact that the "same SPA session" is retained throughout the user action. After we transition into fetching the authorization code, this assumption no longer holds, but instead the application is "hard refreshed" at least for a single time.

Within these changes, this behaviour has been managed with the help of `GdprAuthorizationManager`, `useActionResumer` and `useAuthorizationCode`.

### `GdprAuthorizationManager`

This class is responsible for compliance with the `OpenID` protocol. It's responsible for handling the authorization flow. In this capacity it:
* Stores the application state so that it can be reused when authorization is complete
* Creates the authorization url
* Navigates to authorization url
* Interjects the authorization callback
* Saves token for use
* Reloads application state
* Deletes token and application state from store when it is no longer needed

### `useActionResumer`

This hook is an abstraction which seeks to bridge the "gap" that forms when the user is redirected to Tunnistamo and then finally back into our application. It allows other code within the application to complete actions that span a page refresh  while being relatively agnostic about the method by which the application knows what action to resume when the redirection is done.

**Parameters** 
`deferredAction` - Name of action that get _deferred_  until the redirect back into the UI. This is used as an ID which tells `useActionResumer` whether it should run the `callback` parameter.
`onActionInitialization` - `useActionResumer` begins the action flow by calling this function
`callback` - `useActionResumer` calls this function when the action is resumed

**Humanized explanation**
`useActionResumer` provides its user with the `startAction` function. This function can be used to invoke an action that persists over page reloads. `useActionResumer` listens with a `useEffect` in order to notice when it should complete an action. When it determines that it's a suitable time, it invokes `callback`.

Currently `useActionResumer` relies on a search parameter to know whether it should invoke a `callback`. The `useEffect` that it uses to determine when an action should be completed is not hooked up to listen to changes in location. This way `useActionResumer` won't work unless the search parameter invoking it is present when the component using it is mounted. If the search parameter becomes available after component mount, the callback won't be invoked. Again from another perspective: `useActionResumer` uses the global `location` object to determine whether a search parameter is present. This means that it won't react to location changes.

Using the global location is a sort of anti-pattern which would make it more risky to transition this application into a server rendered application for instance.

### `useAuthorizationCode`
Combines `GdprAuthorizationManager` and `useActionResumer` into a single API that's easier to consume. Code that needs access to an `authorization_code` can hook up to one with a call like this:

```
  const [startFetchingAuthorizationCode, isAuthorizing] = useAuthorizationCode(
    'useDownloadProfile',
    handleAuthorizationCodeCallback
  );
```

## Technical flow

Here I've explained how the application should act in more technical detail. Developers can make use of this explanation to get a better sense of how the features relying on `authorization code` should work. I'll take the download flow as an example, but the delete flow is mostly the same.

1) User logs in
2) User expands panel for downloading user profile
3) User clicks download button
    1) Download button is disabled and its label is changed
    1) Current url and the download action are saved into local storage under `kuvaGdprAuthManager` prefix that's tailed by a random UUID
    1) Tunnistamo authorize URL is built
    1) User is redirected to tunnistamo
6) User allows access to personal information in tunnistamo
7) User is redirected back into the application into address <origin>/gdpr-callback
    1) This view is styled to mimic the general structure of the application by rendering three boxes that correspond in size and colour to the navigation, page heading and content within the profile index page. This makes the juggling of pages less jarring to the user.
    1) It calls `GdprAuthorizationManager.authorizationTokenFetchCallback`
        1) Token is saved into localStorage ready for consumption.
        1) Previous app state is restored. User is redirected to the page the invoked download on and a special search parameter is added to tell `useActionResumer` instances whether they should fire their callbacks.
        1) Previous app state is cleared
8) User lands back on the profile index page based on the redirect.
    1) The expandable element containing the download button is expanded by default.
    1) The page is scrolled so that the download buttons is in view.
    1) The download button is disabled.
    1) The search parameter denoting which action should be fired is removed.
    1) The token is consumed--it's requested from `GdprAuthorizationManager` which then clears it from its own memory.
    1) The token is used to call `downloadProfile`
    1) Once the request completes, the button is enabled again and the file should download.

## Acceptance testing (dev pov)

I've configured the PR environment so that you should be able to use it to complete these acceptance tests:
https://open-city-profile-ui-qa-r-feature-om-8-r1e4rk.test.kuva.hel.ninja/

**EDIT:** It seems that the cert misconfiguration in the infrastructure can lead to a `CERT_ERROR`. You can resolve this error by accessing the API for profiili directly (https://open-city-profile-qa-r-feature-om-8-fe4ln5.test.kuva.hel.ninja/graphql/), and accepting the special dialog Firefox presents.

Before
- Create an account and login
- Open browser console and find panel containing information about `localStorage` (in Chrome Application > Local Storage (on the sidebar on the left) > url of the instance you are testing)

**Downloading profile**  
1) Keep note of local storage--expect to see item with `kuvaGdprAuthManager` once you begin the next step--it's not necessary to try and validate its content
1) Click on the download button
    * Expect to see `kuvaGdprAuthManager` in `localSotrage`
    * Expect button to be disabled
    * Expect to be redirected into Tunnistamo
    * Expect to maybe have to allow permissions
    * Expect to be redirected back to profiili ui
2) Once you are redirected back, expect to land on `/ gdpr-callback` page
    * Expect to see three squares stacked on top of each other reminiscent of the app layout
    * Expect to be redirected into `/` with a search parameter
    * Expect `authorization_code` to be added into `localStorage` (if you are quick enough)
3) Once you are redirected into `/``
    * Expect page to be scrolled to the download button
    * Expect the expandable element containing the download button to be open
    * Expect the download button to be disabled briefly
    * Expect the download to being and complete
    * Expect button to be enabled again
    * Expect the search params to be cleared
    * Expect `localStorage` to be cleared of items beginning with `kuvaGdprAuthManager`

**Deleting profile**

Same as above, but once you are redirected back in step 3, you are redirected again into the deletion page with a timer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+
+## [Unreleased]
+### Added
+- Support for authorization code generation for GDPR API related calls (profile download and deletion) [#108](https://github.com/City-of-Helsinki/open-city-profile-ui/pull/108)
+
+## [1.0.0-rc.1]
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ The graphql-backend for development is located at https://profiili-api.test.kuva
 
 ## Learn More
 
+To learn more about specific choices in this repository, you can browse the [docs](/docs).
+
 You can learn more in the [Create React App documentation](https://facebook.github.io/create-react-app/docs/getting-started).
 
 To learn React, check out the [React documentation](https://reactjs.org/).
--- a/docs/gdpr-api-authorization.md
+++ b/docs/gdpr-api-authorization.md
@@ -1,0 +1,80 @@
+# GDPR API compatibility
+
+The GDPR API requires the user to allow actions on their data. Let's take downloading profile data as an example.
+
+1) User clicks the download button
+2) User is redirected to Tunnistamo which, if necessary, renders an UI the user can use to allow a set of permissions
+3) User is redirected back to Helsinki profile UI and the download action is completed
+
+In essence, we need to request the authorization code in the UI, because the user flow may contain a step that requires user input. Once this code is generated, it must be provided within the download profile query and delete profile mutation when these requests are sent to the profile backend. The backend can then use this code to make requests to all the other services for data or deletion.
+
+## Technical explanation
+
+This flow introduces one difficult step--the exit and re-entry into the profile UI application. This makes the download and deletion code flows more difficult to handle. In comparison, these actions were previously completed with callbacks and promises--which make use of the fact that the "same SPA session" is retained throughout the user action. After we transition into fetching the authorization code, this assumption no longer holds, but instead the application is "hard refreshed" at least once.
+
+Within this application, this behaviour has been managed with the help of `GdprAuthorizationCodeManager`, `useActionResumer` and `useAuthorizationCode`.
+
+### `GdprAuthorizationCodeManager`
+
+This class is responsible for compliance with the `OpenID` protocol. It's responsible for handling the authorization flow. In this capacity it:
+* Stores the application state so that it can be reused when authorization is complete
+* Creates the authorization url
+* Navigates to authorization url
+* Interjects the authorization callback
+* Saves code for use
+* Reloads application state
+* Deletes code and application state from store when it is no longer needed
+
+### `useActionResumer`
+
+This hook is an abstraction which seeks to bridge the "gap" that forms when the user is redirected to Tunnistamo and then finally back into our application. It allows other code within the application to complete actions that span a page refresh while being relatively agnostic about the method by which the application knows what action to resume when the redirection is done.
+
+**Parameters**  
+
+| Name  | Description |
+| ------------- | ------------- |
+| **`deferredAction`**  | Name of action that get _deferred_  until the redirect back into the UI. This is used as an ID which tells `useActionResumer` whether it should run the `callback` parameter.  |
+| **`onActionInitialization`**  | `useActionResumer` begins the action flow by calling this function.  |
+| **`callback`**  | `useActionResumer`** calls this function when the action is resumed.  |
+
+**Humanized explanation**  
+`useActionResumer` provides its user with the `startAction` function. This function can be used to invoke an action that persists over page reloads. `useActionResumer` listens with a `useEffect` in order to notice when it should complete an action. When it determines that it's a suitable time, it invokes `callback`.
+
+**Note:**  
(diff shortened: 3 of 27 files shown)


src/gdprApi/GdprAuthManager.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  clear(key: string) {
+    localStorage.removeItem(`${PREFIX}.${key}`);
+  }
+
+  consumeToken(): string {
+    const token = this.token;
+
+    if (!token) {
+      throw Error('There was no token do it could not be consumed');
```
badrange: "There was no token to consume" maybe?

badrange (commented):
Is there some way to run this locally or in a review env?

Jokinen:
I’ve tested this myself in the review environment. I configured it to work with these changes. It’s accessible here: 

https://open-city-profile-ui-qa-r-feature-om-8-r1e4rk.test.kuva.hel.ninja/

I also wrote down instructions for testing (end of description) which may be helpful.

src/gdprApi/GdprAuthManager.ts
Jokinen: I updated it 👍

tuovinensanttu (approved):
Code is looking good 👍 PR environment did not work for me so I couldn't test this new functionality.

badrange:
The explanation is really well written, it should be documented somewhere for future reference, with a link from the README.md file?

charn:
Had to do some minor hacks to backend to test this (like momentarily ignore SSL certificate checks for GDPR API calls that the backend makes), but I was able to successfully download data for a profile that also has a youth profile and also delete a profile which also has a youth profile. So afaik this could be merged together with https://github.com/City-of-Helsinki/open-city-profile/pull/197 :shipit: 

P.S. going to remove the hack now.

Jokinen:
Big thanks to @charn for getting this thing tested! I'm going to clean up these commits and remove the PR specific configs in order to make this PR ready for merge.

PR #111, opened 2020-06-26
OM-884 | Fix focus indicator being partially hidden

Jokinen:
Screenshot 2020-06-26 at 9 30 00


tuovinensanttu (approved):
Tested UI, focus is wrapped around component beautifully!

PR #113, opened 2020-06-29
Refactor notifications

Jokinen:
After this refactor, multiple notification can be shown at the same
time. Previously notifications would render on top of each other. When
two identical notifications were rendered on top of each other,
dismissing the first would reveal the next, creating the illusion that
closing notification did not work.

Because these notifications were used like toasts, and because other
applications in the office have used this naming, I used the name
toast.


tuovinensanttu (approved):
Read the code, nothing to comment on. Tested the UI as well, notifications are appearing in a column. Nice work! 👍

PR #114, opened 2020-06-29
OM-869 | Add link back to account management

Jokinen:
Refactors expandable panel to not have bottom margin--this way I could avoid setting bottom margin for this new element, and instead set up a grid in `<ProfileInformation />` which allows a single source of truth for the margin amount.

The urls, and the Helsinki account authentication method references are configurable. I documented these in the readme. All of the new variables have sensible defaults, but at least `REACT_APP_IPD_MANAGEMENT_URL_HELSINKI_ACCOUNT` may need to be configured differently based on environment.

Helsinki account still won't be enabled outside dev environments for a while, but that should not make a difference for the UI as it won't show a link back to Helsinki account unless it is told to do so.


tuovinensanttu (approved):
Code is neat and UI redirected me nicely to GH. One nitpicky thingy `profileInformationAccountManagementLink.module.css` should be capitalized.

PR #115, opened 2020-07-01
OM-886 | Replace select with dropdown

tuovinensanttu:
Replace custom `select` component with HDS `Dropdown`


src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -479,13 +497,24 @@ function EditProfileForm(props: Props) {
...
-                            as={Select}
                             name={`addresses.${index}.countryCode`}
                             id={`addresses.${index}.countryCode`}
                             options={countryOptions}
-                            labelText={t('profileForm.country')}
+                            label={t('profileForm.country')}
+                            default={
+                              formikProps.values.addresses[index].countryCode
+                            }
+                            // onChange option type needs to match HDS that's why its set to OptionType | OptionType.
```
badrange: I guess you mean `OptionType | OptionType[]` 

Suggested even shorter comment: 

// onChange is set to OptionType | OptionType[] in order to match hds typing

badrange (approved):
It looks good to me. Good timing for this, we need to do something similar to `kukkuu-ui` at some point. 

Some of the onChange functions are a bit tricky to read, but I can't think of any way to make them simpler or more generic.

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: Changed comments to what you suggested. But yeah, the `onChange` is kinda tricky, because of the typing it's hard to make it more generic than this.

Jokinen: You could also consider if you wanted to force the typing to always be one or the other. The reason for why the types are like so is that the HDS dropdown supports multiselect. Your implementation does not require it. You could hard code your implementation to be "single select" by not allowing the user to define the flag that turns the control into a multiselect. After that you could only have `OptionType` safely. You could likely enforce that type in the in the formik binding component.

You could then, when you need it, define a multi select binding for formik later on.

Jokinen (approved):
Looks good to me 👍

src/profile/components/editProfileForm/EditProfileForm.tsx
tuovinensanttu: Setting `multiselect` to false actually did the trick. Thanks!

PR #125, opened 2020-12-08
OpenShift compatibility

iivorait:
Improved compatibility with OpenShift


Dockerfile
```diff
@@ -63,12 +62,20 @@ COPY . /app
 RUN yarn build
 
 # =============================
-FROM nginx:1.17 as production
+FROM rhel8/nginx-116 as production
```
vikoivun: This is problematic, as it becomes impossible to build the image outside OpenShift. Would be it okay to use explicit URL here? Ie:
registry.redhat.io/rhel8/nginx-116
That would allow building elsewhere as well.

.prod/nginx.conf
```diff
@@ -1,22 +1,39 @@
-server {
-    listen       80;
-    server_name  localhost;
+pid        /tmp/nginx.pid; #for running as non-root 
 
-    location /healthz {
-        return 200 'OK';
-    }
+events {
+    worker_connections  4096;  ## Default: 1024
```
vikoivun, line 4: What is the purpose of this change? Looks benign, but would be nice to know why.

vikoivun (changes requested):

Dockerfile
iivorait: Changed this

.prod/nginx.conf
iivorait: Probably came from a config template for high traffic websites: https://stackoverflow.com/questions/56723449/nginx-error-1024-worker-connections-are-not-enough

Sukriva: Does this need to be changed or can it be left as is?

tituomin: Can be left as is.

tituomin (approved):

tituomin (approved):

vikoivun (approved):

PR #126, opened 2020-12-11
HP-380 fix vulnerabilities

NikoHelle:
Fixes vulnerabilities found by Snyk: https://app.snyk.io/org/city-of-helsinki/project/166fa581-33c0-4bfc-90ba-513e2bf8bf94/

This PR updates vulnerable packages which have an updated version. Some updates were major version changes, so quite many changes had to be made after the update. But mostly fixes for linting errors raised by react-scripts update. Two test files needed actual code changes to fix tests.

Sadly this PR wont fix all vulnerabilities because there are no updates to address new vulnerabilities in updated packages.


.gitignore
```diff
@@ -21,3 +21,4 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+.eslintcache
```
NikoHelle, line 24: .eslintcache is added by react-scripts v4

package.json
```diff
@@ -81,11 +78,13 @@
   },
   "devDependencies": {
     "apollo": "^2.21.1",
+    "eslint": "^7.15.0",
```
NikoHelle: Added because so many dependencies require it on their own.

src/common/header/userDropdown/UserDropdown.tsx
```diff
@@ -12,6 +12,7 @@ import useToast from '../../../toast/useToast';
 
 const NAME_QUERY = loader('../../../profile/graphql/NameQuery.graphql');
 
+// eslint-disable-next-line @typescript-eslint/ban-types
 type Props = {};
```
NikoHelle: These ban-types are removed in another commit.

src/auth/__tests__/authService.test.js
```diff
@@ -25,9 +24,10 @@ describe('authService', () => {
 
   describe('getToken', () => {
     it('should get API_TOKENS from localStorage', () => {
+      const getSpy = jest.spyOn(Storage.prototype, 'getItem');
       authService.getToken();
 
-      expect(localStorage.getItem).toHaveBeenNthCalledWith(3, API_TOKEN);
+      expect(getSpy).toHaveBeenNthCalledWith(1, API_TOKEN);
```
NikoHelle, line 30: It does not makes sense why getItem would be called 3 times in this test. Maybe previous spy did not reset properly between tests.

src/toast/__tests__/Toast.test.jsx
```diff
@@ -44,7 +44,7 @@ describe('Toast', () => {
   });
 
   afterAll(() => {
-    global.document.getElementById('toast').remove();
+    global.document.getElementById('toast-root').remove();
   });
```
NikoHelle, line 48: This test did not make sense either. There is no element with id 'toast'. Previous version of JSDOM may have worked more like jQuery and returned callable functions when query does not return anything. But now calling remove() threw and error.

Also there is only one test, so unsuccessful removal in afterAll() have no effect to following tests.

src/toast/Toast.tsx
```diff
@@ -6,13 +6,20 @@ import { ToastTypes } from './types';
 interface Props {
   title?: string;
   description?: string;
+  // eslint-disable-next-line react/no-unused-prop-types
   id: string;
```
akikoskinen: Is `id` needed at all here, if it's not used?

akikoskinen (approved):
Looks good. One small question.

src/toast/Toast.tsx
NikoHelle: It was needed due the way Toast.tsx is used: by passing Toast object which has the id. This change is now outdated, because I changed the component to use Toast interface.

akikoskinen:
Still looking good.

PR #129, opened 2020-12-22
DEVOPS-30 Dynamically creating client to KuVa Tunnistamo

Hi-Fi:
Jira: DEVOPS-30


public/env-config.js
```diff
@@ -0,0 +1,13 @@
+window._env_ = {
+  REACT_APP_APPLICATION_NAME: "$npm_package_name",
```
Jokinen: Would this work? I've assumed that the `$` syntax is some kind of syntax sugar that's applicable in `.env` files and apparently in Gitlab CI files. I would guess that it does not work in actual `.js` files. One way to circumvent is to import the package object and then find the field.

public/env-config.js
```diff
@@ -0,0 +1,13 @@
...
+  REACT_APP_OIDC_AUTHORITY: "https://api.hel.fi/sso-test/",
+  REACT_APP_OIDC_CLIENT_ID: "https://api.hel.fi/auth/helsinkiprofile-ui",
+  REACT_APP_PROFILE_AUDIENCE: "https://api.hel.fi/auth/helsinkiprofile",
+  REACT_APP_PROFILE_GRAPHQL: "https://profiili-api.test.kuva.hel.ninja/graphql/",
+  REACT_APP_OIDC_SCOPE: "openid profile https://api.hel.fi/auth/helsinkiprofile",
+  REACT_APP_SENTRY_DSN: "https://8b5b23e2171b42cd8617e2b1ad7353b6@sentry.hel.ninja/63",
+  REACT_APP_VERSION: "$npm_package_version",
```
Jokinen: See previous comment.

Jokinen (commented):
The review environment build has failed so the feature can't be tested.

It's unclear how devs should set up a test environment after this change. The `window._env_` is hardcoded to use production settings. It would seem that devs, who tend to develop with the `yarn start` command for speed, would end up starting the UI against the production version of the API. They can edit the `public/env-config.js` file to change what variables are provided, but that seems like an unsafe approach considering that then those changes can end up being merged by accident.

It seems that a script has been set up for reading `.env` files into the Docker container. The `public/env-config.js` file is rebuilt according to `.env` and gets saved into the container. This logic would impose the same downside as mentioned in the previous paragraph when we consider the development experience--changes in the source could accidentally end up in version control.

One more note that bares making is that this application, as CRA applications in general, rely on `dotenv` logic for providing the environment. Although I am personally of the mind that the fewer files the better, some projects, including this one, use variants of the `.env` file like `.env.local` or `.env.development`. These more specific files are mostly used while developing. This is to say that the support for development mode does not achieve feature parity by only tackling the `.env` file.

public/env-config.js
Hi-Fi: Can be fixed/removed. This was just directly from the original .env -file put here as template. Didn't removed yet as didn't know it the value is really needed or just some thing that came with CRA.

Same applies to REACT_APP_VERSION.

public/env-config.js
Hi-Fi: > The review environment build has failed so the feature can't be tested.
> 
> It's unclear how devs should set up a test environment after this change. The `window._env_` is hardcoded to use production settings. It would seem that devs, who tend to develop with the `yarn start` command for speed, would end up starting the UI against the production version of the API. They can edit the `public/env-config.js` file to change what variables are provided, but that seems like an unsafe approach considering that then those changes can end up being merged by accident.
> 
> It seems that a script has been set up for reading `.env` files into the Docker container. The `public/env-config.js` file is rebuilt according to `.env` and gets saved into the container. This logic would impose the same downside as mentioned in the previous paragraph when we consider the development experience--changes in the source could accidentally end up in version control.
> 
> One more note that bares making is that this application, as CRA applications in general, rely on `dotenv` logic for providing the environment. Although I am personally of the mind that the fewer files the better, some projects, including this one, use variants of the `.env` file like `.env.local` or `.env.development`. These more specific files are mostly used while developing. This is to say that the support for development mode does not achieve feature parity by only tackling the `.env` file.

During the yarn start same shell script can be called that would generate the correct file, or some more sophisticated method be used if need to support also Windows.

This approach is not depending on the build time arguments, but the `env-config.js` is generated at the start time of the pod, and it uses variables from environment by default (like e.g. https://github.com/City-of-Helsinki/open-city-profile-ui/pull/129/files#diff-ebc21ccd816b27c07389e15ad88c9e74b71bde5e8e315e4e30902097d38386e5R40-R43). `.env` file is needed to tell the needed values for app, and default values for those.

One big thing to remember is that CRA apps relies `.env` only at build time. At runtime there's no environment variables available at all, as code is not running on the environment where we can set environment variables.

Jokinen: Ok, so I think I misunderstood at least two things: the default window environment is actually the test environment (1). This environment is overridden when the Docker image is started instead of when it is built (2).

> During the yarn start same shell script can be called that would generate the correct file, or some more sophisticated method be used if need to support also Windows.

Hmn, this won't cause as much chafe as I thought because I misunderstood 1. The integration with the `yarn start` command would likely be a shared problem between projects and a common solution could be warranted.

>One big thing to remember is that CRA apps relies .env only at build time. At runtime there's no environment variables available at all, as code is not running on the environment where we can set environment variables.

I am aware that CRA embeds the environment by default. I'm not sure what the point you are making id. Do you mean to give an argument for why the window object needs to be used? I personally think it's a fitting approach in this scenario.

Hi-Fi:
Review environment configured to use Tunnistamo dynamically, and Tunnistamo that environment is also configured to work.

tuovinensanttu (approved):
Logging in with tunnistamo works. LGTM 👍

tituomin (approved):

PR #131, opened 2021-01-05
DEVOPS-202/runtime configuration

Hi-Fi:
Jira: DEVOPS-202


.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+public/env-config.js
```
Jokinen, line 7: It seems that this file is included in the repo, but at the same time ignored here. Is this an accident or by design?

package.json
```diff
@@ -56,13 +56,14 @@
     "yup": "^0.27.0"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "NODE_ENV=development npm run update-runtime-env && react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "NODE_ENV=test npm run update-runtime-env && react-scripts test",
```
Jokinen: I'll try to humanize the changes that `NODE_ENV` goes through when this script is called.

1. `process.env.NODE_ENV` is set to `test`
2. `update-runtime-env` is invoked which saves the current `NODE_ENV` into the `env-config.js` file
3. CRA's `test` script is invoked, which sets `process.env.NODE_ENV` to `test` again
4. During the test command the `setupTests.ts` file is executed which adds the environment from `env-config` into the window object
5. The application uses the environment from window and other tooling may use the environment in the environment variables

I'm not sure, but I could imagine that this would be hard to reason for someone who is familiar with CRA, but not with the repo. I wonder if we should support such devs' understanding with comments (etc)?

public/env-config.js
```diff
@@ -0,0 +1,13 @@
...
+  REACT_APP_PROFILE_GRAPHQL: 'https://helsinkiprofile.test.kuva.hel.ninja/graphql/',
+  REACT_APP_HELSINKI_ACCOUNT_AMR: 'helusername',
+  REACT_APP_APPLICATION_NAME: 'open-city-profile-ui',
+  REACT_APP_OIDC_AUTHORITY: 'https://api.hel.fi/sso/openid/',
+  REACT_APP_OIDC_CLIENT_ID: 'https://api.hel.fi/auth/helsinkiprofile-ui',
+  REACT_APP_PROFILE_AUDIENCE: 'https://api.hel.fi/auth/helsinkiprofile',
+  REACT_APP_OIDC_SCOPE: 'openid profile https://api.hel.fi/auth/helsinkiprofile',
+  REACT_APP_SENTRY_DSN: 'https://8b5b23e2171b42cd8617e2b1ad7353b6@sentry.hel.ninja/63',
+  REACT_APP_VERSION: '1.1.0'
+}
```
Jokinen, line 12: Missing newline

src/setupTests.ts
```diff
@@ -8,7 +8,9 @@ const customGlobal: GlobalWithFetchMock = global as GlobalWithFetchMock;
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 customGlobal.fetch = require('jest-fetch-mock');
 customGlobal.fetchMock = customGlobal.fetch;
-
+global.window = {
+    _env_: require('../public/env-config').window
```
Jokinen: The `env-config` file in this changeset has the following structure `window._env_`. Shouldn't this line then result in a window object like this:

```js
window: {
  _env_: {
    _env_ : { /* ... */ }
  }
}
```

Jokinen: I'm not entirely sure what the `.window` does after the `require`. Doesn't the `env-config` file just mutate the `window` object?

.gitignore
Hi-Fi: Accident, I thought I made also the removal.

public/env-config.js
Hi-Fi: Adding also .editorconfig that handles these.

src/setupTests.ts
Hi-Fi: That was the way tests were passing. Have to debug those to see how jest makes it's magic.

package.json
Hi-Fi: `react-scripts test` implicitly sets NODE_ENV to test. Issue is, that at that time there already needs to be the valid configuration for test, and to create that env needs to be given.

NODE_ENV is used in the code, so it has to go to the file (at serviceWorker and authService)

Jokinen (commented):
Nice documentation!

I gave this a whirl locally and I noticed one more thing that's worth pointing out. Because the same file is used for every process, the environment is in essence shared between processes. If I `start` the application in one terminal, and then `test` in another, the application process will end up with `NODE_ENV` set to `test` as well. Not all devs will run the two commands at the same time, but for instance CRA sort fo promotes this approach by always starting tests in watch mode locally. To get everything back to normal the dev would need to edit the environment source manually or restart the `start` script.

Some applications do check against `NODE_ENV === development` which could then theoretically unexpectedly fail. Although I would still expect for such breakages to be relatively uncommon in practice.

Jokinen:
One more comment, sorry for the spam. Because Travis no longer works for us linting wasn't checked, but it seems there are some error level problems. These may fail in the CI later on once it works again.

src/setupTests.ts
Hi-Fi: Thought error, require itself was enough to create the object that was used in tests.

Hi-Fi:
> Nice documentation!
> 
> I gave this a whirl locally and I noticed one more thing that's worth pointing out. Because the same file is used for every process, the environment is in essence shared between processes. If I `start` the application in one terminal, and then `test` in another, the application process will end up with `NODE_ENV` set to `test` as well. Not all devs will run the two commands at the same time, but for instance CRA sort fo promotes this approach by always starting tests in watch mode locally. To get everything back to normal the dev would need to edit the environment source manually or restart the `start` script.
> 
> Some applications do check against `NODE_ENV === development` which could then theoretically unexpectedly fail. Although I would still expect for such breakages to be relatively uncommon in practice.

Thank you, I have already an idea how to handle this (create configs for test with different name if that value is set which already defines environment to `test` by default. Will also check the lint part and probably remove travis as it's quite stuck again.

Hi-Fi:
> One more comment, sorry for the spam. Because Travis no longer works for us linting wasn't checked, but it seems there are some error level problems. These may fail in the CI later on once it works again.

Splitted CI fix to DEVOPS-210

scripts/update-runtime-env.ts
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env ts-node-script
+
```
Jokinen, line 2: Lots of stuff in this file that doesn't seem complete. Commented out imports, missing spaces and `;`. Pointing out so it isn't forgotten.

scripts/update-runtime-env.ts
```diff
@@ -0,0 +1,42 @@
...
+const defaultNodeEnv = process.env.TEST ? 'test' : 'development'
+// Prevent collision is app is running while tests are started
+const configFile = process.env.TEST ? 'test-env-config.js' : 'env-config.js'
+
+process.env.NODE_ENV=process.env.NODE_ENV || defaultNodeEnv
+
+const getClientEnvironment = require('../node_modules/react-scripts/config/env.js');
+
+
+const condigurationFile: string = path.join(__dirname, '../public/'+configFile);
```
Jokinen: Typo in variable name.

Jokinen (approved):
I tested out running the testing process in parallel and it looked to be working now.

As the `update-runtime-env` is a JS script, we have the option to share it by using https://github.com/City-of-Helsinki/helsinki-utils-js once we start to propagate this approach, which is nice.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
NikoHelle, line 12: This looks like specific editor settings? Those should not be included.

README.md
```diff
@@ -53,6 +64,15 @@ Fetches translation data from our Google Spreadsheet and updates translation fil
 
 You still need to update tests and add the translation files to the git repository manually.
 
+### `yarn update-runtime-env`
+
+Generetes variable object used when app is running. Generated object is stored at `public/env-config.js` and available
```
NikoHelle: Tiny typo in "Generetes"

NikoHelle:
One thing might be worth mentioning. I recently updated react-scripts to v4 in this repo. That included a change in the order which .env -files are merged. 

Before env.develop / .env.production took precedence over .env.local. Now .local has higher priority.

https://github.com/facebook/create-react-app/pull/9037/files

.editorconfig
Hi-Fi: It's not: https://editorconfig.org/

I'd think it would help with consistency of code no matter what editor or IDE is used.

Hi-Fi:
> One thing might be worth mentioning. I recently updated react-scripts to v4 in this repo. That included a change in the order which .env -files are merged.
> 
> Before env.develop / .env.production took precedence over .env.local. Now .local has higher priority.
> 
> https://github.com/facebook/create-react-app/pull/9037/files

Code uses directly react-scripts env-file loading, so if the is not changed it shouldn't affect to this (meaning would work like plain react-scripts)

scripts/update-runtime-env.ts
Hi-Fi: Cleaned file up

NikoHelle (approved):

PR #132, opened 2021-01-11
HP-42 service view update

NikoHelle:
Changes made according to UX-design in https://helsinkisolutionoffice.atlassian.net/browse/HP-404. KuVa had already made all features.

Just small CSS and translation changes. Only FI-translations exist, so EN and SV have placeholders for translators.

Currently there are no BE end points for removing or getting data of only one service. Therefore links "Poista tietosi tästä palvelusta" / "Lataa kaikki tietosi" were not added.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -79,7 +79,7 @@
   "nav": {
     "information": "Information",
     "menuButtonLabel": "Profile menu",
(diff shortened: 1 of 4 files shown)


akikoskinen (approved):
Looks good to me

PR #133, opened 2021-01-12
HP-385 fix bugs and code smells reported by Sonar

NikoHelle:
Original list of bug, security hotspots and code smells: https://sonarcloud.io/dashboard?id=City-of-Helsinki_open-city-profile-ui

Done outside of this PR:
- excluded all generated files from checks
- deactivated rule for  "Shorthand object properties should be grouped at the beginning or end of an object declaration"

### DIFF
--- a/src/auth/authService.ts
+++ b/src/auth/authService.ts
@@ -84,13 +84,14 @@ export class AuthService {
   }
 
   public async login(path = '/'): Promise<void> {
-    try {
-      return this.userManager.signinRedirect({ data: { path } });
-    } catch (error) {
+    let success = true;
+    await this.userManager.signinRedirect({ data: { path } }).catch(error => {
+      success = false;
       if (error.message !== 'Network Error') {
         Sentry.captureException(error);
       }
-    }
+    });
+    return success ? Promise.resolve() : Promise.reject();
   }
 
   public async endLogin(): Promise<User> {
--- a/src/profile/components/editProfileForm/EditProfileForm.tsx
+++ b/src/profile/components/editProfileForm/EditProfileForm.tsx
@@ -36,22 +36,24 @@ import FormikDropdown, {
   HdsOptionType,
 } from '../../../common/formikDropdown/FormikDropdown';
 
+const maxLengthValidation = 'validation.maxLength';
+
 const addressSchema = yup.object().shape({
-  address: yup.string().max(128, 'validation.maxLength'),
-  city: yup.string().max(64, 'validation.maxLength'),
-  postalCode: yup.string().max(5, 'validation.maxLength'),
+  address: yup.string().max(128, maxLengthValidation),
+  city: yup.string().max(64, maxLengthValidation),
+  postalCode: yup.string().max(5, maxLengthValidation),
 });
 
 const phoneSchema = yup.object().shape({
   phone: yup
     .string()
     .min(6, 'validation.phoneMin')
-    .max(255, 'validation.maxLength'),
+    .max(255, maxLengthValidation),
 });
 
 const schema = yup.object().shape({
-  firstName: yup.string().max(255, 'validation.maxLength'),
-  lastName: yup.string().max(255, 'validation.maxLength'),
+  firstName: yup.string().max(255, maxLengthValidation),
+  lastName: yup.string().max(255, maxLengthValidation),
   language: yup.string(),
   primaryPhone: phoneSchema,
   primaryAddress: addressSchema,
--- a/src/profile/components/editProfileForm/EditProfileForm.tsx
+++ b/src/profile/components/editProfileForm/EditProfileForm.tsx
@@ -101,12 +101,10 @@ function EditProfileForm(props: Props): React.ReactElement {
 
   const applicationLanguage = getLanguageCode(i18n.languages[0]);
   const countryList = countries.getNames(applicationLanguage);
-  const countryOptions = Object.keys(countryList).map(key => {
-    return {
-      value: key,
-      label: countryList[key],
-    };
-  });
+  const countryOptions = Object.keys(countryList).map(key => ({
+    value: key,
+    label: countryList[key],
+  }));
 
   const getFieldError = (
     formikProps: FormikProps<FormValues>,
@@ -148,12 +146,10 @@ function EditProfileForm(props: Props): React.ReactElement {
   };
 
   const profileLanguageOptions: OptionType[] = profileConstants.LANGUAGES.map(
-    language => {
-      return {
-        value: language,
-        label: t(`LANGUAGE_OPTIONS.${language}`),
-      };
-    }
+    language => ({
+      value: language,
+      label: t(`LANGUAGE_OPTIONS.${language}`),
+    })
   );
   return (
     <Formik
@@ -225,8 +221,11 @@ function EditProfileForm(props: Props): React.ReactElement {
                 options={profileLanguageOptions}
                 default={formikProps.values.profileLanguage}
                 label={t('profileForm.language')}
-                onChange={(option: HdsOptionType) =>
-                  formikProps.setFieldValue('profileLanguage', option.value)
+                onChange={option =>
+                  formikProps.setFieldValue(
+                    'profileLanguage',
+                    (option as HdsOptionType).value
+                  )
                 }
               />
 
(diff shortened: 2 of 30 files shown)


src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -12,6 +12,9 @@ import { FormValues } from '../../components/editProfileForm/EditProfileForm';
 import { updateMutationVariables } from '../updateMutationVariables';
 import { myProfile } from '../../../common/test/myProfileQueryData';
 
+const constantTestEmail = 'kolmas@testi.fi';
+const constantTestAddress = 'Testikatu 66';
```
akikoskinen: In my opinion this change (the whole commit bbcca29d3c) doesn't improve the code. When looking at for example this test code change, I think the previous version is better. The [Sonar rule](https://sonarcloud.io/organizations/city-of-helsinki/rules?open=typescript%3AS1192&rule_key=typescript%3AS1192) is very simplistic and doesn't consider the context at all. As such I would even consider disabling the whole rule.

src/profile/components/editProfileForm/EditProfileForm.tsx
```diff
@@ -185,7 +185,7 @@ function EditProfileForm(props: Props): React.ReactElement {
       }}
       validationSchema={schema}
     >
-      {(formikProps: FormikProps<FormValues>) => (
+      {(editProfileFormikProps: FormikProps<FormValues>) => (
```
akikoskinen: This commit (8fb9f7a5ca) doesn't look to me like the right solution. The tool is helpful, it's indicating a code smell -- it says it's code duplication, but it could also be too large components, wrong architecture or a number of other things. Instead of silencing the tool by making worse code, find a solution that **improves** the code. That's anyway the idea in using these tools in the first place.

src/auth/authService.ts
```diff
@@ -85,12 +85,13 @@ export class AuthService {
...
     try {
-      return this.userManager.signinRedirect({ data: { path } });
+      await this.userManager.signinRedirect({ data: { path } });
     } catch (error) {
       if (error.message !== 'Network Error') {
         Sentry.captureException(error);
       }
-      return Promise.reject();
+    } finally {
+      Promise.resolve();
```
akikoskinen: If error occurs, previously a rejected Promise was returned, now a resolved Promise. That changes the behavior of this function. Is that ok?

src/subscriptions/components/subsciptions/Subscriptions.tsx
```diff
@@ -132,7 +132,7 @@ function Subscriptions(): React.ReactElement {
   const handleCheckboxValues = (
     index: number,
     code?: string,
-    value?: boolean
+    value = false
```
akikoskinen, line 135: Can this still also have the explicit type? Something like this perhaps:

    value: boolean = false

akikoskinen (changes requested):
These one-liner commit messages are really ugly, especially with complex URLs within them. Use line breaks in commit messages.

Screenshot 2021-01-15 at 14 33 56

src/subscriptions/components/subsciptions/Subscriptions.tsx
NikoHelle: No, it raises an error. Setting a default value also types it. "value" can only be true / false after that.

Error: "Type boolean trivially inferred from a boolean literal, remove type annotation."

src/profile/components/editProfileForm/EditProfileForm.tsx
NikoHelle: Yes, agreed. I removed the "fix" and the code smell can remain. These components that have duplicate code will be refactored anyway when user data from strong authentication is added. Better to fix this then.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Fixed properly by using the values from variables where values are initially set.

src/subscriptions/components/subsciptions/Subscriptions.tsx
akikoskinen: Got it.

src/auth/authService.ts
NikoHelle: login() causes a full page redirect and none of the callers listen to the result. But returned value should of course be logical. I made some changes to catch a possible error and return rejected / resolved promise in a simpler way.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -134,7 +134,7 @@ test('add arrays are formed correctly', () => {
 
   expect(variables.input.profile.addAddresses).toEqual([
     {
-      address: 'Testikatu 66',
+      address: addresses[1].address,
       postalCode: '00000',
       city: 'Helsinki',
       countryCode: 'FI',
```
akikoskinen: There's still a lot of repetition even if the tool doesn't complain anymore. Would it be possible to use the `addresses[1]` as a whole (perhaps selecting only wanted fields out of it or something)?

akikoskinen (changes requested):

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: The original test seems to be done quick & dirty. There's so much repetition. I've now commited massive changes to the tests. They are not squashed yet, because if it is not satisfactory yet, I don't want to squash multiple times.

akikoskinen: Dosn't look that massive to me. But cleaner 😃

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -97,33 +112,12 @@ const formValues: FormValues = {
...
-    id: '123',
-    primary: true,
-    address: 'Testikatu 55',
-    city: 'Helsinki',
-    countryCode: 'FI',
-    postalCode: '00100',
-    addressType: AddressType.OTHER,
-    __typename: 'AddressNode',
-  },
+  primaryEmail: cloneObject(emails[0]),
```
akikoskinen: Could this clone the `primaryEmail`?

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -97,33 +112,12 @@ const formValues: FormValues = {
...
+  emails,
+  phones,
+  primaryPhone: cloneObject(phones[0]),
```
akikoskinen: Could this clone the `primaryPhone`?

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -97,33 +112,12 @@ const formValues: FormValues = {
...
+  addresses,
+  primaryAddress: cloneObject(addresses[0]),
```
akikoskinen: Could this clone the `primaryAddress`?

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -175,30 +162,13 @@ test('update arrays are formed correctly', () => {
   );
 
   expect(variables.input.profile.addAddresses).toEqual([
```
akikoskinen: Not related to this PR in any way, but should this be checking a field named `updateAddresses`?

akikoskinen (approved):
Few comments still, but looks good already like this. I suppose solving the conflict with the current `develop` branch doesn't change the contents of this PR in any way.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Tests might break, if data not in the arrays (emails, addresses, phones) is used here where test data is created.  "primaryEmail/Phone/Address" are currently used in the array, but that could change or their data could be used with modifications. 

Bottom line: If arrays change, the changes should not break the tests.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Same comment as above

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Same comment as above

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Good catch. That made no sense and is now fixed.

akikoskinen (approved):
Pretty nice.

PR #140, opened 2021-02-04
Merge previous 1.1.2 hotfixes into master

charn:
Earlier 1.1.2 hotfixes were deployed using github flow branching model. These fixes were removed when a new version from master was deployed. This PR merges those changes into master.


Jokinen (approved):
This looks good to me.

PR #142, opened 2021-02-15
Resource requests and limits set

Hi-Fi:
Jira: DEVOPS-231


Jokinen (commented):
Based on our discussion on Slack, we should maybe double the CPU limit for production based on the data in Grafana.

Jokinen (approved):

PR #146, opened 2021-03-17
HP-318 step 1: Update HDS

NikoHelle:
HDS-library needed to be updated for new layouts. New version has many components that were also in Profile UI, so replaced them from HDS.

HDS Notification component uses react-spring library which has a bug, so made a copy of it without react-spring. Basically just took out animations.

The bug:
pmndrs/react-spring#1078


akikoskinen:
Some test is failing?

NikoHelle:
> Some test is failing?

Ah, one snapshot was not updated.

src/common/notification/NotificationComponent.tsx
```diff
@@ -18,16 +19,17 @@ function NotificationComponent(props: Props): React.ReactElement | null {
   }
   return (
     <div className={styles.notification}>
-      <DismissableNotification
+      <Notification
+        dismissible
```
akikoskinen, line 23: Probably something I don't know about modern React, but how does this `dismissible` line work? Does the value somehow come from `props` or what's going on here?

akikoskinen: Ahh, now I remember. It's that same as `dismissible=true`, right?

NikoHelle: Yeah! And linter actually forbids writing dismissible={true}, so it must be a plain "dismissible" which looks like an undefined variable.

akikoskinen (approved):
LGTM

PR #147, opened 2021-03-18
HP-318 step 2: GraphQL update

NikoHelle:
Apollo client was old and deprecated and did not support cache updating. Changed to client to new one - from same authors so switch was easy.

Updated MyProfileQuery to get Verified Personal Information and render it with a new component. Added suomi.fi as type of authorisation where needed.


akikoskinen:
@NikoHelle can you rebase on top of the current `develop` branch?

NikoHelle:
> @NikoHelle can you rebase on top of the current `develop` branch?

Done!

yarn.lock
```diff
@@ -3424,6 +3452,11 @@ autoprefixer@^9.6.1:
     postcss "^7.0.32"
     postcss-value-parser "^4.1.0"
 
+await-handler@^1.1.2:
+  version "1.1.2"
+  resolved "https://registry.yarnpkg.com/await-handler/-/await-handler-1.1.2.tgz#329c19e9382599898bc12c80c47efdcdaa1b4992"
+  integrity sha512-dihteGhwbJpT89kVbacWiyKeAZr+En0YGK6pAKQJLR0En9ZxSH2H4TTvfG4bBjzFq9gDAma4y9BrpDns6j5UiQ==
+
```
akikoskinen: `await-handler` dependency is added in commit 475885465bb...

yarn.lock
```diff
@@ -3452,11 +3452,6 @@ autoprefixer@^9.6.1:
     postcss "^7.0.32"
     postcss-value-parser "^4.1.0"
 
-await-handler@^1.1.2:
-  version "1.1.2"
-  resolved "https://registry.yarnpkg.com/await-handler/-/await-handler-1.1.2.tgz#329c19e9382599898bc12c80c47efdcdaa1b4992"
-  integrity sha512-dihteGhwbJpT89kVbacWiyKeAZr+En0YGK6pAKQJLR0En9ZxSH2H4TTvfG4bBjzFq9gDAma4y9BrpDns6j5UiQ==
-
```
akikoskinen: ...`await-handler` dependency is removed in commit f8f631bc8a80. And that's the only contents of this commit. Can commit 475885465bb be changed so that `await-handler` dependency is not added (apparently it's not needed)? And commit f8f631bc8a80 can thus be removed.

.env.development
```diff
@@ -1,3 +1,5 @@
-REACT_APP_OIDC_AUTHORITY="https://api.hel.fi/sso-test/"
-REACT_APP_OIDC_CLIENT_ID="https://api.hel.fi/auth/helsinkiprofile-ui"
-REACT_APP_PROFILE_GRAPHQL="https://profiili-api.test.kuva.hel.ninja/graphql/"
+REACT_APP_OIDC_AUTHORITY="https://tunnistamo.dev.hel.ninja/"
+REACT_APP_OIDC_CLIENT_ID="https://api.hel.fi/auth/helsinkiprofile-ui-dev"
+REACT_APP_PROFILE_AUDIENCE="https://api.hel.fi/auth/helsinkiprofiledev"
+REACT_APP_PROFILE_GRAPHQL="https://helsinki-profile-api-dev.agw.arodevtest.hel.fi/graphql/"
+REACT_APP_OIDC_SCOPE="openid profile https://api.hel.fi/auth/helsinkiprofiledev"
```
akikoskinen, line 5: Is this (and the corresponding `public/env-config.js` change) somehow related to the apollo client update, as these changes are now in the same commit? I don't understand the relevance.

src/graphql/generatedTypes.ts
```diff
@@ -298,8 +301,100 @@ export interface MyProfileQuery_myProfile_phones {
...
+  readonly temporaryAddress: MyProfileQuery_myProfile_verifiedPersonalInformation_temporaryAddress | null;
+  /**
+   * The temporary foreign (i.e. not in Finland) residency address.
+   */
+  readonly permanentForeignAddress: MyProfileQuery_myProfile_verifiedPersonalInformation_permanentForeignAddress | null;
+}
+
 export interface MyProfileQuery_myProfile {
-  readonly __typename: "ProfileNode";
+  readonly __typename: "ProfileWithVerifiedPersonalInformationNode";
```
akikoskinen: In a [very recent version of open-city-profile](https://github.com/City-of-Helsinki/open-city-profile/commit/9fdfb228de57117a362ff8051ae18097197ca718), the type returned by `myProfile` query is once again called `ProfileNode`. Perhaps calling it temporarily `ProfileWithVerifiedPersonalInformationNode` is not needed in this UI code.

src/i18n/en.json
```diff
@@ -135,16 +161,37 @@
...
+    "noEmails": "You haven't given any emails",
+    "noAddresses": "You haven't given any addresses"
   },
   "sanityCheck": "open-city-profile.en",
   "serviceConnections": {
     "clock": "at",
     "created": "Created:",
     "empty": "You currently don't have any services connected to your profile.",
-    "explanation": "EN: Olet antanut seuraaville palveluille oikeuden käyttää Helsinki Profiili -tietojasi.",
+    "explanation": "You have given the following services permission to use you Helsinki Profile data.",
```
akikoskinen: Typo: second "you" should probably be "your".

src/i18n/en.json
```diff
@@ -135,16 +161,37 @@
...
+    "primaryAddress": "This is your primary address",
+    "permanentAddress": "EN: Vakinainen osoite",
+    "temporaryAddress": "EN: Väliaikainen osoite",
+    "permanentForeignAddress": "EN: Väliaikainen ulkomainen osoite",
+    "primaryEmail": "This is your primary email",
+    "primaryPhone": "This is your primary phone number",
+    "verifiedDataInformationPrefix": "EN: Merkityt",
+    "verifiedDataInformationAriaPrefix": "EN: Alla olevat",
+    "verifiedDataInformation": "EN: <0></0> tiedot on vahvistettu virallisesta väestötietojärjestelmästä, eikä niitä ei voi tämän takia muokata Helsinki Profiilissa. Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: <1>siirry suomi.fi palveluun </1>",
+    "verifiedDataInformationLink": "https://testipalvelu.apro.tunnistus.fi",
```
akikoskinen, line 177: I don't know whether this URL needs to be localizable. But at least it needs to be configurable in different deployment environments. I have no idea how to best achieve that, though.

src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
```diff
@@ -0,0 +1,197 @@
...
+        ]}
+      />
+    </p>
+  );
+
+  return (
+    <ProfileSection hasVerifiedUserData>
+      <h3
+        className={commonFormStyles.sectionTitle}
+        aria-label="Vahvistetut perustiedot"
```
akikoskinen: Should `aria-label`s be localized?

src/profile/helpers/getCountry.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import getLanguageCode from '../../common/helpers/getLanguageCode';
+import { MyProfileQuery_myProfile_addresses_edges_node as AddressNode } from '../../graphql/generatedTypes';
+
+export default function getCountry(
+  countryCode: AddressNode['countryCode'] | undefined,
+  lang: string
+): string {
+  return countries.getName(countryCode || 'FI', getLanguageCode(lang));
+}
```
akikoskinen, line 11: Ooooh, internationalized country names. Very nice 👍

src/config.ts
```diff
@@ -32,6 +32,10 @@ const config = {
     window._env_.REACT_APP_IPD_MANAGEMENT_URL_YLE,
     'https://tunnus.yle.fi/#omat-tiedot'
   ),
+  identityProviderManagementUrlTunnistusSuomifi: defaultTo(
+    window._env_.REACT_APP_IPD_MANAGEMENT_URL_TUNNISTUS_SUOMI_FI,
```
akikoskinen, line 36: It seems these configurations have been mentioned in the README. Perhaps this new one should be added there too.

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -4,34 +4,50 @@ import { loader } from 'graphql.macro';
...
 import ConfirmationModal from '../modals/confirmationModal/ConfirmationModal';
 import ExpandingPanel from '../../../common/expandingPanel/ExpandingPanel';
 import Button from '../../../common/button/Button';
-import { ServiceConnectionsQuery } from '../../../graphql/generatedTypes';
+import { ServiceConnectionsQuery as ServiceConnectionsRoot } from '../../../graphql/generatedTypes';
 import useToast from '../../../toast/useToast';
 import styles from './deleteProfile.module.css';
+import useDeleteProfile from '../../../gdprApi/useDeleteProfile';
+import checkBerthError from '../../helpers/checkBerthError';
+import BerthErrorModal from '../modals/berthError/BerthErrorModal';
```
akikoskinen, line 18: I wonder what all these \*berth\* things are... But I see things were just moved around, so no need to do anything about that in this PR. But perhaps need to write it down somewhere that there could be something that should be checked if it's still correct (as there are and will be other services than just berth).

src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
akikoskinen: Aha, it **is** localized, but in another commit.

akikoskinen (changes requested):
I would expect there to be some tests as there is new functionality.

.env.development
NikoHelle: Previous, very old, configs did not provide a login flow that results in getting Verified Personal Information. Time and need to update them to use OpenShift.

src/graphql/generatedTypes.ts
NikoHelle: This is an autogenerated file from a script that gets its data from graphQL-server. Script hasn't been run for a while. I'll run it again and reflect changes to the codebase.

src/i18n/en.json
NikoHelle: Yes, there is no way to make environment dependant translations at the moment. It should be done when there is actual need. Maybe all links can just point to production versions until other versions are actually required. This url was given to me, because suomi.fi was apparently wrong.

src/profile/components/deleteProfile/DeleteProfile.tsx
NikoHelle: Good point!

yarn.lock
NikoHelle: Later on, in next PR, await-hander is needed and will be added. I removed it from package.json for this PR, but forgot to run "yarn" to remove it from yarn.lock. But then tests fail, because yarn.lock is not in sync and "yarn install" in test container is trying to create new yarn.lock without write privileges. So I had to update yarn.lock.

src/i18n/en.json
NikoHelle: Done. This is update is from somewhere else... But all translation files are updated simultaneously.

src/graphql/generatedTypes.ts
NikoHelle: Done.

yarn.lock
NikoHelle: f8f6 is now removed. await-handler was removed in a new commit that updated yarn.lock anyway and await-handler in not seen anymore as a change.

src/config.ts
NikoHelle: Done.

.env.development
akikoskinen: But could the config be changed in its own commit, if it's not tightly related to the apollo client update?

NikoHelle: Sliced and diced the file (+ auto-generated env-config.js) to another commit.

NikoHelle:
> I would expect there to be some tests as there is new functionality.

Tests are done, but added in the next PR because they depend on libraries and data mocking that are not part of Apollo update.

akikoskinen:
> > I would expect there to be some tests as there is new functionality.
> 
> Tests are done, but added in the next PR because they depend on libraries and data mocking that are not part of Apollo update.

That makes it a bit hard to do a proper review as tests and production code always go hand in hand. Now I can't see the other hand.

NikoHelle:
> > > I would expect there to be some tests as there is new functionality.
> > 
> > 
> > Tests are done, but added in the next PR because they depend on libraries and data mocking that are not part of Apollo update.
> 
> That makes it a bit hard to do a proper review as tests and production code always go hand in hand. Now I can't see the other hand.

New functionalities are in components that only render text and 2 links: VerifiedPersonalInformation and  ProfileInformationAccountManagementLink. No new logic. That's why tests are not included here. I can copy tests from next PR, but it requires a lot of work getting them isolated from other tests and exclude the code not needed in this PR.

NikoHelle:
> > > > I would expect there to be some tests as there is new functionality.
> > > 
> > > 
> > > Tests are done, but added in the next PR because they depend on libraries and data mocking that are not part of Apollo update.
> > 
> > 
> > That makes it a bit hard to do a proper review as tests and production code always go hand in hand. Now I can't see the other hand.
> 
> New functionalities are in components that only render text and 2 links: VerifiedPersonalInformation and ProfileInformationAccountManagementLink. No new logic. That's why tests are not included here. I can copy tests from next PR, but it requires a lot of work getting them isolated from other tests and exclude the code not needed in this PR.

src/graphql/generatedTypes.ts
akikoskinen: Could squash the fixing commit to the one that introduces the `ProfileWithVerifiedPersonalInformationNode` names in the first place.

yarn.lock
akikoskinen: `await-handler` is still added to, and then removed from `yarn.lock`, in this PR. But `await-handler` is not in `package.json` in any commit in this PR. I don't understand why it has to be in `yarn.lock` for a few commits. What brings it there?

NikoHelle: yarn.lock is an autogenerated file and should not be edited by hand. await-handler was previously in package.json, so it is also here. This commit was edited afterwards. But lock file do not change unless yarn in also run. Running yarn separately just for this commit is pointless.

NikoHelle: All yarn.lock updates are now reset from commits and created a new commit just for new yarn.lock.

akikoskinen: > yarn.lock is an autogenerated file and should not be edited by hand.

Yes, that's clear.

> await-handler was previously in package.json, so it is also here.

What do you mean by "was previously"? Running `git log -S await-handler` while this PR's branch is checked out, gives no results, so it's not found anywhere in the commit history.

> This commit was edited afterwards. But lock file do not change unless yarn in also run. Running yarn separately just for this commit is pointless.

No. Yarn (or whatever package management tool is used) **must** be run every time the dependency list file (e.g. `package.json`) is changed. And the lock file **must** be included in the same commit. If yarn was not run for a commit where `package.json` was changed, you use interactive rebase to go back to that commit that modified `package.json`, run yarn and rewrite the commit adding the changed lock file.

> All yarn.lock updates are now reset from commits and created a new commit just for new yarn.lock.

That's wrong. `Package.json` and related `yarn.lock` changes need to be in the same commit. You can rewrite history as long as this is a PR we are talking about.

NikoHelle: await-handler cannot be found in commit history, because that change in package.json is not in any commit anymore. That caused this discussion as it was in one commit and removed in another. 

yarn.lock in now back in each commit that alters package.json.

src/graphql/generatedTypes.ts
NikoHelle: This has been done yesterday.

akikoskinen (approved):

PR #150, opened 2021-03-24
HP-318 step 3: centralise code

NikoHelle:
Centralise typings, schemas and form properties. Also form validation and creating error messages is now simpler and does not require to define data multiple times in different places.

### DIFF
--- a/src/graphql/typings.ts
+++ b/src/graphql/typings.ts
@@ -1,0 +1,102 @@
+import {
+  MyProfileQuery,
+  MyProfileQuery_myProfile_verifiedPersonalInformation,
+  MyProfileQuery_myProfile,
+  MyProfileQuery_myProfile_addresses,
+  MyProfileQuery_myProfile_addresses_edges,
+  MyProfileQuery_myProfile_addresses_edges_node,
+  MyProfileQuery_myProfile_emails,
+  MyProfileQuery_myProfile_emails_edges,
+  MyProfileQuery_myProfile_emails_edges_node,
+  MyProfileQuery_myProfile_phones,
+  MyProfileQuery_myProfile_phones_edges,
+  MyProfileQuery_myProfile_phones_edges_node,
+  UpdateMyProfile,
+  UpdateMyProfile_updateMyProfile_profile,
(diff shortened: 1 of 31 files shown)


src/graphql/typings.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export type PermanentForeignAddress = MyProfileQuery_myProfile_verifiedPersonalInformation_permanentForeignAddress;
+export type PermanentAddress = MyProfileQuery_myProfile_verifiedPersonalInformation_permanentAddress;
+export type TemporaryAddress = MyProfileQuery_myProfile_verifiedPersonalInformation_temporaryAddress;
+
+export type ServiceConnectionsRoot = ServiceConnectionsQuery;
+export type ServiceConnectionsNode = ServiceConnectionsQuery_myProfile_serviceConnections_edges_node;
+export type Service = ServiceConnectionsQuery_myProfile_serviceConnections_edges_node_service;
+// eslint-disable-next-line max-len
+export type ServiceAllowedFieldsNode = ServiceConnectionsQuery_myProfile_serviceConnections_edges_node_service_allowedDataFields_edges_node;
```
NikoHelle, line 86: Eslint-disable could be avoided by:

type SAFN = ServiceConnectionsQuery_myProfile_serviceConnections_edges_node_service_allowedDataFields_edges_node;
export type ServiceAllowedFieldsNode = SAFN;

Which is not an improvement IMHO

akikoskinen: Is 

```
export type ServiceAllowedFieldsNode =
ServiceConnectionsQuery_myProfile_serviceConnections_edges_node_service_allowedDataFields_edges_node;
```

syntax error or somehow otherwise unallowed?

NikoHelle: Linter moves it back into one line.

akikoskinen: So linter first makes too long lines and then complains about them? I've encountered the same situation 😄 Sometimes they are hard to please. This is good as it is.

akikoskinen (approved):

PR #151, opened 2021-03-24
HP-318 step 4: centralised data handling and querying

NikoHelle:
## Centralised code for data storing, reading and updating

Profile data is stored to one context and that context is provided to all components with React Context provider. Getting and updating data is now done in one place and used easily with React hooks. 
Data mutations are done with single mutation editor.

Updates with UpdateMyProfile-query are also stored to cache instead of re-querying data with MyProfile query again. This required update to the query-file, now it returns full profile after update.


src/profile/helpers/mutationEditor.ts
```diff
@@ -0,0 +1,520 @@
...
+    return (profileDataItem as PhoneNode).primary === true;
+  }
+  if (dataType === 'emails') {
+    return (profileDataItem as EmailNode).primary === true;
+  }
+  if (dataType === 'addresses') {
+    return (profileDataItem as AddressNode).primary === true;
+  }
+  return false;
+}
```
akikoskinen, line 249: So far I have only been looking at the first commit in this PR, which adds this one file. This comment is about the whole file, although I marked this one function within the file.

I find this whole module doing things in a convoluted way. There are many functions which take in some data and a data type. Then within the function body there's a long `if..else if` structure that checks the data type and does something according to it.

Typescript supports polymorphism and that's exactly a tool to avoid these kinds of long `if..else if`s. As an example, this `isPrimary` function. There could for example exist an interface, let's call it `Primaryable` (please come up with a better name). It has one accessor (field, function, whatever you have available in TS), `primary`. The phone, email and address types implement the `Primaryable` interface. Now you can call `data.primary` for any of these types directly and this `isPrimary` function isn't needed at all anymore.

Repeat the same pattern for other "features" these data types support, and I think you can get rid of most of the functions in this module.

akikoskinen (changes requested):

src/profile/helpers/mutationEditor.ts
NikoHelle: You mean:
```
interface Data = {
data:<"some MyProfileNode">
}

interface Primaryable extends Data {
isPrimary():boolean
setPrimary():Primaryable
}

class Address implements Primaryable {
isPrimary(){return this.data.primary}
}
```
"isPrimary" could also be "primary" and a const.

akikoskinen: For example something like that. Or if Typescript supports implementing multiple interfaces (I don't know if it does), then `Primaryable` shouldn't extend `Data`, but `Address` would implement both `Data` and `Primaryable`.

NikoHelle: This pattern is for OOP and mainly classes. Current code is functional and there are no classes or objects that have own functions and access to "this". 

In functional programming (FP) just pass the object to a function (usually pure) and it returns what is needed. 

FP: isPrimary(data); // (x)=>return x.primary
OOP: data.isPrimary(); // ()=>return this.primary;

In OOP, instance must also be created to have "isPrimary" and "this". With FP, need for neither exists.

I prefer functional programming over classes. One function handles all scenarios, instead of making multiple functions to each instance. More verbose, but easier to test. Typescript makes it more verbose as it does not allow returning objectX.primary as a boolean if all possible objectXs dont have primary, like basicData.

I can add new interfaces that might save a few lines of code and if-spaghetti.

NikoHelle:
I'll make smaller commit and PR, so closing this one.

PR #152, opened 2021-04-06
HP-318: Step 4 - new profile data handling

NikoHelle:
This PR has new tools for profile data fetching and storing. Also few hooks for components for easier data usage.  And a lot of testing tools. Functionalities for data editing and mutating are in next PRs.

Introducing ProfileContext, a React context as single source of ProfileData accessible from all components.  Also added  first version of editData that will be used for editing profile data in all editors (basic data, phones, addresses, etc) and hooks.


src/common/test/MockApolloClientProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+    : undefined;
+  return !errorType
+    ? {
+        data,
+      }
+    : errorType === 'networkError'
+    ? { error: new Error(`NetworkError at ${Date.now()}`) }
+    : {
+        errors: [new GraphQLError(`GraphQLError at ${Date.now()}`)],
+      };
```
akikoskinen: Quite a many ternary operators here. I find it somewhat hard to read. Changing some of them to `if..elses` could IMO improve readability.

```suggestion
  if (!errorType) {
    const data = profileData
      ? { myProfile: profileData }
      : { updateMyProfile: { profile: updatedProfileData } };
    return { data };
  } else if (errorType === 'networkError') {
    return { error: new Error(`NetworkError at ${Date.now()}`) };
  } else {
    return {
      errors: [new GraphQLError(`GraphQLError at ${Date.now()}`)],
    };
  }
```

src/profile/context/ProfileContext.tsx
```diff
@@ -0,0 +1,173 @@
...
+  data?: ProfileRoot;
+  updateProfileData: (data: ProfileRoot) => void;
+  error?: ApolloError | Error;
+  loading: boolean;
+  addErrorListener: (listener: ErrorListener) => ErrorDisposer;
+  fetch: () => void;
+  refetch: () => Promise<QueryResult>;
+  isInitialized: boolean;
+  isComplete: boolean;
+  getName: (preferNickOrGivenName?: boolean) => string;
```
akikoskinen, line 34: The function argument accepts more than one kind of falsy value, (`false` and `undefined`, if the argument is not provided, I suppose). Is it possible to make it accept only proper `boolean` values (so, remove the `?`), but then provide a default value for the argument, so callers can still leave it out?

src/profile/context/__tests__/ProfileContext.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      await errorPromise;
+      context = result.current;
+      expect(context.data).toBeUndefined();
+      expect(context.error).toBeDefined();
+      expect(context.loading).toEqual(false);
+      expect(context.isInitialized).toEqual(true);
+      expect(context.isComplete).toEqual(false);
+      // apollo triggers error listeners later, "waitForTimeout" added 'cause of that
+      await waitForTimeout(100);
+      expect(errorListener.mock.calls.length).toEqual(1);
```
akikoskinen: Suggesting to replace a fixed delay with waiting for a condition:

```suggestion
      await waitFor(() => expect(errorListener.mock.calls.length).toEqual(1));
```

 Needs importing `waitFor` from `@testing-library/react`.

src/profile/context/__tests__/ProfileContext.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      listenerDisposer();
+      const secondErrorPromise = waitForErrorChange();
+      context.refetch();
+      await secondErrorPromise;
+      await waitForTimeout(100);
+      expect(errorListener.mock.calls.length).toEqual(1);
```
akikoskinen: This second part doesn't guarantee that the listener was removed. It could get called a second or two later, but this test doesn't catch that.

I suggest adding two listeners, then removing one of them, and checking (using the `waitFor` tactic proposed above) that the remaining listener got called but the removed one not.

Also this listener removing check could be its own test completely.

src/profile/helpers/editData.ts
```diff
@@ -0,0 +1,202 @@
...
+        : getAddressesFromNode;
+    const nodes: MultiItemProfileNode[] = getter({ myProfile: profileData });
+    const primaryData = getPrimary(profileData, dataType);
+    if (primaryData) {
+      const primaryAsNode = createNewProfileNode<MultiItemProfileNode>(
+        dataType,
+        primaryData
+      );
+      nodes.unshift(primaryAsNode);
+    }
```
akikoskinen: The `getPhonesFromNode`, `getEmailsFromNode` and `getAddressesFromNode` filter the primary member out of the list that they return. And then the primary node is added to the list here again. Is that really necessary to do like that? Coult the `get*FromNode` functions be changed to **not** filter the primary node away?

src/profile/helpers/editData.ts
```diff
@@ -0,0 +1,202 @@
...
+  | PhoneValue
+  | EmailValue;
+
+export type FormValues = {
+  firstName: string;
+  nickname: string;
+  lastName: string;
+  primaryEmail: PrimaryEmail;
+  primaryAddress: PrimaryAddress;
+  primaryPhone: PrimaryPhone;
```
akikoskinen, line 67: The primary nodes are included in the `addresses`, `emails` and `phones` arrays received via GraphQL. Could it simplify the code if these `primary*` members wouldn't be included also as separate members? These wouldn't necessarily need to be included in the GraphQL call at all. I don't know if removing these would simplify the code (also the opposite could happen), but I thought to throw this out as an idea to ponder.

src/profile/helpers/editData.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export const basicDataType = 'basic-data';
+export const additionalInformationType = 'additional-information';
+
+export type EditDataType =
+  | 'phones'
+  | 'emails'
+  | 'addresses'
+  | typeof basicDataType
+  | typeof additionalInformationType;
```
akikoskinen, line 52: Why don't the "phones", "emails" and "addresses" strings have constants, like `basicDataType` and `additionalInformationType` have? Instead those strings are present in many places in the code base, waiting for an accidental typo to happen.

src/profile/helpers/editData.ts
```diff
@@ -0,0 +1,202 @@
...
+  emails: EmailNode[];
+  phones: PhoneNode[];
+};
+
+type SaveType = 'value' | 'set-primary' | 'remove' | undefined;
+
+export type EditData = {
+  readonly id: string;
+  readonly value: EditDataValue;
+  readonly primary?: boolean;
```
akikoskinen, line 79: Not sure why a `primary` property is needed here -- it's not (yet?) used anywhere in the code. But, for those types of data (the multi item types) that have a primary concept, have the primary property within the `value` member. Why does it has to be repeated here at the top level of `EditData` too?

src/profile/components/profile/Profile.tsx
```diff
@@ -52,16 +69,18 @@ function Profile(): React.ReactElement {
...
   }, [checkProfileExists, history]);
 
-  const isLoadingAnything = Boolean(isCheckingAuthState || loading);
-  const isProfileFound = data && data.myProfile;
-  const hasGraphQLError = error
-    ? !parseGraphQLError(error).isAllowedError
-    : false;
+  const isDoingProfileChecks = Boolean(isCheckingAuthState || loading);
+  const isProfileFound = !!(data && data.myProfile);
+  const hasGraphQLError = error && error.graphQLErrors.length;
```
akikoskinen: Three different ways to create a boolean(ish) value on three consecutive lines. Some consistency could be useful.

src/index.css
```diff
@@ -72,4 +72,5 @@
 
   --hds-primary-font: var(--font-helsinki-grotesk);
   font-family: var(--font-helsinki-grotesk);
+  --color-black-ultra-light: #fafafb;
```
akikoskinen, line 75: Almost white is indeed a very very light black :smile:

src/profile/hooks/useProfileDataEditor.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!profileData) {
+    throw new Error(
+      'No profile data for useProfileDataEditor. Hook used before data has been fetched.'
+    );
+  }
+  const { getEditData } = useMemo(
+    () => createEditorForDataType(profileData, dataType),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
```
akikoskinen, line 29: Shouldn't `profileData` and `dataType` be listed as dependencies here?

src/profile/context/ProfileContext.tsx
```diff
@@ -0,0 +1,173 @@
...
+    loadTracker.current.complete = true;
+    updateProfileData(data);
+  } else if (!profileData && data && !loading) {
+    loadTracker.current.started = true;
+    loadTracker.current.complete = true;
+    loadTracker.current.loading = false;
+    updateProfileData(data);
+  }
+
+  const contextData: ProfileContextData = {
```
akikoskinen, line 118: React contexts weren't that familiar to me, so I went and studied about them. I learned that context changes trigger redraws. So I guess, in order to avoid needless redraws, the context shouldn't seem to change when it doesn't really change. One way to cause unnecessary context changes, i.e. redraws, is to always create a new object for the context. It seems to me that this happens here.

Indeed, the [React documentation](https://reactjs.org/docs/context.html#caveats) warns about not creating a new object each time. Although the example there uses a class and its state, I think the main point holds with functions too: don't create new context objects when not needed.

But when the context is an object (like it is here), and the context (its contents) **does** change, is the only way to notify the context consumers by creating a new context object? Or is it enough to only change the object's members that have changed? I don't know.

Does this raise any ideas?

src/profile/context/ProfileContext.tsx
```diff
@@ -0,0 +1,173 @@
...
+};
+
+export const useProfileListener = (listener: ErrorListener): void => {
+  const { addErrorListener } = useContext(ProfileContext);
+
+  const disposeListener = useMemo(
+    () => addErrorListener(listener),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
```
akikoskinen: Related to the React context comment, my study began here. I thought that `addErrorListener` and `listener` should be dependencies to the `useMemo` call. So I investigated why they aren't. It seems to me that it's because they change all the time, even though they don't _really_ change. They change because new functions are provided all the time, but those functions do exactly the same thing as the previous ones. So those functions doesn't _necessarily_ need to be new. The old ones could be used as well.

So. Would the correct implementation be so that `addErrorListener` and `listener` _are_ listed as dependencies for the `useMemo`, but they would only change when they _really_ change? Which in this case is probably never (during the lifetime of the component). And that would result into pretty much the same functionality, but this `useMemo`  would have the dependencies listed correctly.

src/profile/components/profile/__tests__/Profile.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    cleanup();
+    resetApolloMocks();
+  });
+
+  it('should re-route to /login when user is not authenticated', async () => {
+    const { waitForElement } = await renderTestSuite([]);
+    await waitForElement({ testId: 'location-is-login' });
+  });
+
+  it('should render load indicator and then CreateProfile when profile does not exist', async () => {
```
akikoskinen, line 56: I can't see how this test checks for any load indicator, as the test name suggests.

akikoskinen (commented):
Some comments, mainly for opening discussions.

src/common/test/MockApolloClientProvider.tsx
NikoHelle: Changed ternary jungle into separate functions

src/profile/context/ProfileContext.tsx
NikoHelle: Sonar also marks these boolean "double definitions" as code smells, but in case of an arrow function, adding a default value will result in compiler error.

This is ok:

const getName1 = (preferNickOrGivenName = false):string=>{return preferNickOrGivenName ? 'a':'b'}; 


Sadly, this is not:

export type ProfileContextData = {
  getName: (preferNickOrGivenName = false) => string;
};

Error: A parameter initializer is only allowed in a function or constructor implementation.
Same happens in latest TS version in https://www.typescriptlang.org/play?ts=4.3.0-beta#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgBSgnQQBtgBhCFUGAEUxkzjgG8AoFhAZwEkkEMBJlIIAXsAAmALjgAjCBHKYkAbk6JuVALZhyqWQqXAV6lgHNgMAHKZtwWQAowUYOjTWE2ANYB5KADiCABuwEi29gC88orKSACUcJEAfHDcMFDI5uoAvuqgkLBwMACeYHiExGSU1Kgg9IzM0RwWVhEOcM6u7lCePv5BoeF2eNFYpNzAiSlpGVlmcJY2IwBMTi5uHl5+gSFh7QD80kZx06npmUjZ7Hns2NTpi20jAIxJnRs9fTuD+yPv40m8WkFyyKVYrhgAFcoEg4J8tv1dkN2nADnAAOSYDHSDFyDF5OB3B7wJbtFbvLqbXrbAZ7Yb2I4xYwqYGgq7gyEwuEImlI34MvDorE4vEE1RE9joKFIbBCahPZb2ADMVK+tORf3sAJEQJBcyuEKs3Ph3URP3pqOF2Nx+MJ7GJSEeTUV7Rejni6nuTvgmAp0TJr0cGShUy9JLgmGV70D9ndgLDQA

src/profile/context/__tests__/ProfileContext.test.tsx
NikoHelle: Changed

src/profile/context/__tests__/ProfileContext.test.tsx
NikoHelle: Good catch! Fixing this test resulted in finding a bug in Apollo client: onError listeners are triggered only once. Took some time to find it out and fix it. Link to bug is in code explaining why the work-around was added.

src/profile/helpers/editData.ts
NikoHelle: 😮 Has the myProfileQuery always returned primary nodes in the node list too? The old Profile UI code seems odd in that case. And the old query result used in tests is incorrect. Primary nodes are not in the associated node lists. (Fixed now). 

Profiili API seems to return primary nodes correctly as #0 in node lists. In old code, updateMutationVariables uses primaries, but otherwise they are not needed in Profile UI. ..if they are always the first node in the node list. Is this guaranteed? The query does not seem to support sorting, so UI cannot set it. 

I'd prefer deleting primaryAddress, primaryEmail and primaryPhone in another PR.

I added a new param to all get*FromNode, so EditData doesn't need to play with primaries.

src/index.css
NikoHelle: 😄 HDS colors tokens are named --color-black-05 ---> --color-black-100

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Not in this case. "createEditorForDataType" should never be run again, because it initialises editor data. Currently there is no way to update profile, but in the future updates are done separately.

src/profile/context/ProfileContext.tsx
NikoHelle: Redraws when context changes is a designed feature here. Child components require it. UI is simple and doesn't update even every minute, so it is not a performance issue. A lot of code simplicity is gained with context being single source of truth and passing data, error and loading states to child components

After first fetch, error and loading are never triggered again. So only changes in "data"  cause rerenders. Data is updated when mutations update cache.

"the context shouldn't seem to change when it doesn't really change. "  
There are no meaningless changes. Of course, every child/grand-grand-grandchild component does not need all data.

src/profile/context/ProfileContext.tsx
NikoHelle: Dependencies are not included in the array, because it is important that listener is added exactly once and useMemo should never be run again. Empty dependency array takes care of that.

But "addListener" is now added as dependency, because I changed the code in ProfileContext and it never changes.

"listener" is now also a dependency, because it is passed as a never changing function created with new hook: usePersistFn. It always stores the last version of the listener and makes sure it is called. This way re-render loops or multiple identical listeners cant happen.

src/profile/components/profile/__tests__/Profile.test.tsx
NikoHelle: Added missing line.

src/profile/helpers/editData.ts
NikoHelle: Typescript wont allow typos. Cant use 'adress' anywhere where EditDataType is expected.

basicDataType (and currently unused additionalInformationType) are defined as const because they are used also as strings.

Example in current code:
"const testId = basicDataType;"

(Yes, "const testId:EditDataType = 'basic-data';" is an alternative.) 

"Address" etc are always passed to func/params expecting EditDataType or just passed on as a "dataType" variable.

src/profile/helpers/editData.ts
NikoHelle: Answered better in the other primary-related comments. Perhaps  these are obsolete at the moment? Why were they ever added/needed? 😄

src/profile/helpers/editData.ts
NikoHelle: This is only place where primary is defined.

"editData.value" only has value(s) needed in the input fields. 

Values are {phone:'<phone>'},  {email:'<email>'},  {address:'street', city...} etc.

src/profile/context/ProfileContext.tsx
akikoskinen: Right. No default values in type definitions. Makes sense. So in the type, you have to mark the argument as optional. Then in the functions implementing that type you can use default values for (optional) arguments. So this is possible:

```typescript
type MyType = {
  f: (flag?: boolean) => string;
};

const myValue: MyType = {
  f: (flag: boolean = false): string => {
    return typeof flag + String(flag);
  }
};

console.log(myValue.f());
console.log(myValue.f(true));
console.log(myValue.f(false));
```

That way the argument is always boolean within the function. I don't know what Sonar says about this combination.

src/profile/helpers/editData.ts
akikoskinen: I don't know if the primaries have _always_ been included in the lists. But they have been there at least as long as I have been around.

Currently it's guaranteed in API code that the primary comes first in the list, so it doesn't happen by accident. But I'm not sure if it's a public API guarantee (meaning, there would be tests ensuring it remains so, and the behaviour would be documented). I think we can make it a hard guarantee, documented and all.

src/profile/helpers/editData.ts
akikoskinen: Not sure, but perhaps for such clients that only need e.g. one email address so that they can send some email to the customer. So a short hand to avoid searching through the whole list.

src/profile/context/ProfileContext.tsx
NikoHelle: The function itself does not need to have ":boolean" again.

const myValue: MyType = {
  f: (flag = false): string => {
    return typeof flag + String(flag);
  }
};

is valid. 

Added this to the file.

akikoskinen (changes requested):
Some first impressions, about commit order.

The first commit of this PR (0a94e73eb) uses the `ahooks` library. But that library is added as a dependency only in later commit (af6d2d1044). That makes the first 7 commits in this PR unable to build.

New arguments are provided to the `getPhonesFromNode`, `getEmailsFromNode` and `getAddressesFromNode` functions in commit af1eb5b3f0. The functions themselves get the ability to receive those arguments only in a later commit bc3199b9.

As a result, all but the last commit in this PR are unable to build or some tests fail. I would suggest to reorder and/or edit the commits so that every commit compiles and tests pass. It's beneficial for example to make it possible to use `git bisect` should it ever be needed.

NikoHelle:
Reordered latest changes and ran tests in each commit.

src/profile/helpers/editData.ts
akikoskinen: Right, the `EditDataValue`'s are those limited ones without the `primary` member. I don't quite get why there needs to be these limited types, why not just use the `AddressNode`, `EmailNode` and `PhoneNode` types directly (or include the `primary` member into the limited types). Perhaps I will learn it when the editing capabilities start appearing.

src/profile/hooks/useProfileDataEditor.ts
akikoskinen: So, this hook (`useProfileDataEditor`) implementation makes assumptions, that the caller never changes the `dataType` argument, and neither does the profile data change? I can't say if that's a common way to implement hooks, but feels weird to me.

src/profile/context/ProfileContext.tsx
akikoskinen: Yep. I'm sure this isn't causing any performance issues (at least yet), there are so few redraws.

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Yes, dataType never changes in the same component that uses the hook. Same instance of the component never render first "emails" and then "phones". 

And initialiser (createEditorForDataType) should not be called twice. Unsaved user input would be destroyed. In the layout user may have added a new phone number and 3 others numbers edited, but not saved yet. So saving one and re-creating data, would destroy unsaved ones.

 Each Component (and dataType) uses their own instance of the hook. So it's like creating a class instance on component initialisation. It is encapsulated to that component.

src/profile/helpers/editData.ts
NikoHelle: Primary is not a value that user can write, so that's why it is not in the value-property. And checking changes in the value when user has changed the primary would cause false positives. And value is should be compatible with the form output.

src/profile/context/ProfileContext.tsx
akikoskinen: Good that the dependencies are now there.

But isn't it now (and used to be too) so that the user of the `useProfileErrorListener` hook can't change the listener even if they wanted to? Is it this hook's job to decide that the user can not change the listener afterwards? Shouldn't it be the caller's freedom to decide if they want to change listeners at some point?

I'm perhaps pondering general hook design "guidelines" here, learning how they should generally be implemented. Within a single project like this anything is off course possible, as long as the consequences are understood.

src/profile/hooks/useProfileDataEditor.ts
akikoskinen: Is it this hook's job to say whether the caller changes the `dataType` argument or not? This hook shouldn't care in which kinds of components it's used. The components using this hook are probably responsible for handling data updates properly and if that includes keeping `dataType` constant, then a component is free to do that.

If this **hook** really can't handle changing `dataType` argument, then shouldn't it e.g. throw an exception in case `dataType` changes instead of just not acting on it in any way?

Btw. I've been reading [this long discussion](https://github.com/facebook/create-react-app/issues/6880) about missing dependencies and seems that most of the cases of having empty dependency lists have been debunked.

src/profile/context/ProfileContext.tsx
NikoHelle: Listener changes every time. The function passed to useProfileErrorListener() is always stored over the previous one in usePersistFn(). 

https://github.com/alibaba/hooks/blob/master/packages/hooks/src/usePersistFn/index.ts

But changing it does not trigger the useMemo and useEffect.

Alternative to usePersistFn would be to wrap the listener with "useCallback" where ever useProfileErrorListener() is called from. But usePersistFn simplifies the process and "useCallback" is not required every time and listener can be any kind of function.

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: > Is it this hook's job to say whether the caller changes the dataType argument or not? 

Yes, because the hook has only one purpose. It is not for universal use.

>This hook shouldn't care in which kinds of components it's used. 

It doesn't care. It works on any component without errors. It just prevents re-initialisation the data the hook is using or user's data would be destroyed.

Of course, I can add

```
useMemo(()=>{
 throw new Error("DataType changed")
},[dataType]).
```

It will never trigger, but developer can see that it is not supported - if cant be seen from empty dependency array.

It is empty because it should never be run again and it's much easier to realise it when seeing an empty array. 

There is no situation in the UI where component would switch dataTypes. The component is made for one fixed dataType which is obvious from usage. 

If one day there is a new dataType switcher, the component would not use a hook which purpose is to **keep** data of single dataType.

akikoskinen: The empty dependency list is an implementation detail. The future developer might only check the interface of the hook. They see this:

```typescript
function useProfileDataEditor({
  dataType,
}: {
  dataType: EditDataType;
}): ProfileDataEditorReturnType
```

A function which takes in an `EditDataType` and returns a `ProfileDataEditorReturnType`. Nothing in this interface indicates that the user should always use the same value for the `dataType` argument.

This applies to any programming interface: the API should be intuitive (and sometimes the implementation isn't even available to be read). In this regard I'm thinking that hooks are just functions. Am I thinking this somehow incorrectly?

I'm probably over thinking this now, but could there be a "hook factory"? Something like this:

```typescript
function getUseProfileDataEditor({
  dataType,
}: {
  dataType: EditDataType;
}): () => ProfileDataEditorReturnType {
  return (): ProfileDataEditorReturnType => {
    // The hook implementation would be here...
  };
}
```

That would be used something like this then:

```typescript
// Create hook function somewhere
const useProfileDataEditor = getUseProfileDataEditor({dataType: 'basic-data'});
// ...
// Then use it somewhere...
const { editDataList } = useProfileDataEditor();
```

That would prevent passing different `dataType` arguments to the hook. I'm not sure what should be in the dependency list within the hook in this case :thinking:

NikoHelle: Hooks are not just functions that return a value. They were created so functional components can have state that do not change every time the function is called - which is on every render.

From the docs: "With Hooks, you can extract stateful logic from a component " 

Like React's own **useState**(initialValue), the **useProfileDataEditor**(dataType) returns the same value every time even if the arguments change. Same logic there  - once called they do not change on next calls. They keep the state. 

Hook's inner state is controlled with what ever they return in the function call. useState returns its value and a function to change it. If useProfileDataEditor would allow a new dataType, it would return a function for that. React developers who use hooks must know these rules.

>Nothing in this interface indicates that the user should always use the same value for the dataType argument.
 
Datatype does not have to be the same every time. It just has no effect after the first call. Like in useState(). Hooks are hooks, and not just functions, for a reason - they don't mutate/re-init when called. 

Hooks interface really cannot tell you that what a hook does internally. Interface of useState: 
`function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];`
Return value is important part of the hook's interface.

One main concept with hooks is that they cannot be called conditionally. These will result in compiler error

if(dataType === "phones"){
  usePhonesDataEditor(....)
}

if(!initialData){
 [initialData] = useState(createEditorForDataType(dataType))
}

Second concept is that they must be called inside a component. For these reasons factory functions would not work. 
There could be usePhoneDataEditor, useEmailDataEditor etc. But then also PhoneDataEditor etc that uses the hook.
Singletons or Redux could be other solutions. Or create new contexts: /PhoneEditorDataContext, EmailEditorDataContext....

akikoskinen: Thanks for a thorough explanation -- and reminder, I have read these before, but it seems to require some repetition until I fully get it. The analogy with `useState` makes sense.

akikoskinen (approved):

PR #153, opened 2021-04-09
HP-626: Added expiration check to getUser

NikoHelle:
When user has left session open without a logout and returns, the userManager gets user from the localStorage without checking if the user has expired. After that fetching user's profile will return an error. Fetching profile is useless and now avoided by checking also user.expired instead of just existence of user.


VilleMiekkoja (approved):
Looks good 👍

PR #154, opened 2021-04-18
HP-318: step 5: basic data editing

NikoHelle:
Added a HTML form to BasicData-component and buttons to enter edit mode and save changes. New mutation hook for saving data. Data update functions added to editData. 

A lot of new tests and mock updates.


src/common/test/MockApolloClientProvider.tsx
```diff
@@ -23,8 +25,13 @@ export function MockApolloClientProvider({
   responseProvider: ResponseProvider;
 }): React.ReactElement {
   const customGlobal = global as GlobalWithFetchMock;
-  customGlobal.fetchMock.mockResponse((variables?: RequestVariables) => {
-    const response = createMockedProfileResponse(responseProvider(variables));
+  customGlobal.fetchMock.mockResponse(async (req: Request) => {
```
VilleMiekkoja: Does this mock all fetch calls? Is there a risk that this mock stays on for some tests that would not wish to mock fetch?

src/common/test/componentMocking.tsx
```diff
@@ -81,8 +127,8 @@ export const renderProfileContextWrapper = async (
     });
   };
 
-  const waitForDataChange: TestTools['waitForDataChange'] = async () => {
-    const latest = getLastTime('dataUpdateTime');
+  const waitForDataChange: TestTools['waitForDataChange'] = async previousChangeTime => {
+    const latest = previousChangeTime || getLastTime('dataUpdateTime');
```
VilleMiekkoja, line 131: Minor improvement suggestion to help readability: Could the getLastTime function be defined little closer where it is applied? Seems to be also applied inside submit function, but maybe we could put the definition above this waitForDataChange?

src/common/test/componentMocking.tsx
```diff
@@ -17,26 +22,67 @@ import {
...
+  waitForElementAndValue: (props: WaitForElementAndValueProps) => Promise<void>;
+};
+
+export const cleanComponentMocks = (): void => {
+  cleanup();
+  cleanupHooks();
+  resetApolloMocks();
 };
 
 export const emptyResponseProvider: ResponseProvider = () => ({});
```
VilleMiekkoja, line 88: Off-topic comment for line 90: the renderProfileContextWrapper seems to do more than just rendering; it also provides various test utility functions. Can't figure for a better name though. Maybe it is doing too much?

src/common/test/componentMocking.tsx
```diff
@@ -97,14 +143,33 @@ export const renderProfileContextWrapper = async (
     return waitForDataChange();
   };
 
-  const getElement: TestTools['getElement'] = ({ testId, text }) => {
+  const getElement: TestTools['getElement'] = ({
```
VilleMiekkoja, line 146: It seems that this function only wants one of the arguments per function call? Im not sure if it's easy to tell from the api of the function, that user should only pass one of them? One option could be to separate the qualifier and value. Something like

`const getElementByQualifier(qualifier, value)`. Though then we would need to type all the qualifiers, and that would also have an impact to the readability probably. 🤔

src/common/test/componentMocking.tsx
```diff
@@ -119,17 +184,74 @@ export const renderProfileContextWrapper = async (
...
+      () => {
+        getElement({ valueSelector: newValue });
+      },
+      { timeout: 60 }
+    );
+  };
+
+  const waitForElementAndValue: TestTools['waitForElementAndValue'] = async props => {
+    const { selector, value } = props;
+    const elementValue = await getTextOrInputValue(selector);
```
VilleMiekkoja, line 249: Minor improvement suggestion: The getTextOrInputValue function seems to be applied only inside this function, and also it is inside the return promise. I wonder if we could define it above this waitForElementAndValue function so it would be closer to usage?

src/common/test/componentMocking.tsx
```diff
@@ -119,17 +184,74 @@ export const renderProfileContextWrapper = async (
...
           }
-          value = element.textContent || undefined;
         }
       },
       { timeout: 150 }
     );
     return Promise.resolve(value);
   };
 
+  const triggerAction: TestTools['triggerAction'] = async selector => {
```
VilleMiekkoja: Would clickElement describe this function better?

src/common/cssHelpers/form.module.css
```diff
@@ -139,7 +139,7 @@
```
VilleMiekkoja: dev.hel.fi recommends _camelCase_ for style file names: https://dev.hel.fi/coding-standards#naming

src/common/test/componentMocking.tsx
```diff
@@ -232,6 +354,74 @@ export const exposeProfileContext = (
...
+
+export const ProfileContextFetcher = ({
+  children,
+}: {
+  children: React.ReactElement | React.ReactNodeArray;
+}): React.ReactElement => {
+  const [fetchStarted, setFetchStarted] = useState(false);
+  const { data, fetch } = useContext(ProfileContext);
+  if (!fetchStarted) {
+    fetch();
```
VilleMiekkoja, line 411: Data fetching side effects are usually wrapped inside useEffect hook. Is it not needed this time?

src/common/test/MockApolloClientProvider.tsx
```diff
@@ -23,8 +25,13 @@ export function MockApolloClientProvider({
   responseProvider: ResponseProvider;
 }): React.ReactElement {
   const customGlobal = global as GlobalWithFetchMock;
```
VilleMiekkoja: I see that the jest-fetch-mock library is updated in this pr from 2.1.2 to 3.0.3 version. The documentation of 3.0.3 jest-fetch-mock (https://www.npmjs.com/package/jest-fetch-mock) no longer mention this type _GlobalWithFetchMock_ . So I wonder if the preferred way of handling globals with the jest-fetch-mock library has changed?

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+  return _.get(profileData, targetPath, null);
+}
+
+function getEdgeTypeName(
+  dataType: EditDataType
+): 'EmailNodeEdge' | 'AddressNodeEdge' | 'PhoneNodeEdge' {
+  return dataType === 'emails'
+    ? 'EmailNodeEdge'
+    : dataType === 'phones'
+    ? 'PhoneNodeEdge'
```
VilleMiekkoja: Personally I find double ternary little confusing, but could be just that I haven't got to used to it. Would probably prefer regular if else for clarity

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+
+function setPrimaryNode(
+  profile: Mutable<ProfileData>,
+  dataType: EditDataType,
+  newPrimary: Mutable<MultiItemProfileNode> | null
+): void {
+  if (newPrimary) {
+    newPrimary.primary = true;
+  }
+  const propName = getPrimaryPropName(dataType);
```
VilleMiekkoja, line 275: The getPrimaryPropName function seems to be applied only inside this setPrimaryNode function. It might increase readability if it would be defined straight above this setPrimaryNode function

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+  const currentProfile = _.cloneDeep(profileData) as Mutable<ProfileData>;
+  // added nodes must have empty id in formValues but id in the created GraphQL
+  // which requires same extra check ups and deletion:
+  const addedNodes: Set<string> = new Set();
+  return {
+    getProfile() {
+      return currentProfile;
+    },
+    add(dataType, data, index = -1) {
+      const edge = createEdge(dataType, data);
```
VilleMiekkoja, line 291: The createEdge function seems to be applied only inside this cloneAndManipulateProfile function. Would it increase readability to have it defined straight above this cloneAndManipulateProfile function so it would be closer to usage? Though the setPrimaryNode function is also applied inside this cloneAndManipulateProfile function.

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+): ManipulationFunctions {
+  const currentProfile = _.cloneDeep(profileData) as Mutable<ProfileData>;
+  // added nodes must have empty id in formValues but id in the created GraphQL
+  // which requires same extra check ups and deletion:
+  const addedNodes: Set<string> = new Set();
+  return {
+    getProfile() {
+      return currentProfile;
+    },
+    add(dataType, data, index = -1) {
```
VilleMiekkoja, line 290: This seems to be missing typings?

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+      return this;
+    },
+    setAdditionalInformation(additionalInformation) {
+      Object.assign(currentProfile, additionalInformation);
+      return this;
+    },
+    getFormValues() {
+      const basicData = pickSources(
+        currentProfile,
+        basicDataType
```
VilleMiekkoja, line 345: Does basicDataType refer to first name, last name and nickname? Would _personNamesDataType_ describe it better?

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
...
+  newPrimary: Mutable<MultiItemProfileNode> | null
+): void {
+  if (newPrimary) {
+    newPrimary.primary = true;
+  }
+  const propName = getPrimaryPropName(dataType);
+  Object.assign(profile, { [propName]: newPrimary });
+}
+
+export function cloneAndManipulateProfile(
```
VilleMiekkoja: I find the naming of this function little problematic. It does clone the profile, but does not really give the result without calling the additional functions that it returns. It feels more like a class type of object than a function that does a particular thing. Maybe renaming it to _clonedProfileWithManipulationFunctions_ etc would be more describing?

src/graphql/generatedTypes.ts
```diff
@@ -608,9 +605,42 @@ export interface UpdateMyProfile_updateMyProfile_profile_primaryAddress {
...
   readonly id: string;
+  readonly primary: boolean;
   readonly address: string;
   readonly postalCode: string;
   readonly city: string;
+  readonly countryCode: string;
+  readonly addressType: AddressType | null;
+}
+
+export interface UpdateMyProfile_updateMyProfile_profile_addresses_edges_node {
```
VilleMiekkoja, line 616: The amount of different naming conventions this interface applies is ambitious to say the least 😅 . These graphql types seems to be automatically generated? Could we configure the naming conventions somehow? There also seems to be some repetition with the word _UpdateMyProfile_ that I'm not sure if it provides any additional value?

src/profile/components/basicData/BasicData.tsx
```diff
@@ -1,28 +1,172 @@
...
   const { t } = useTranslation();
+  const {
+    content,
+    setErrorMessage,
+    setSuccessMessage,
+    clearMessage,
+  } = useNotificationContent();
   const testId = basicDataType;
 
   if (!editDataList || !editDataList[0]) {
```
VilleMiekkoja, line 45: This seems to be the first row where return values from _useProfileDataEditor_ function call are being applied. I wonder if it would increase readability if we make the function call _useProfileDataEditor_ straight above this line 45, so that it would be closer to the place where it is used?

src/profile/components/basicData/BasicData.tsx
```diff
@@ -1,28 +1,172 @@
...
+  const { value, saving } = editData;
   const { firstName, nickname, lastName } = value as BasicDataValue;
   const formFields = getFormFields(basicDataType);
 
+  const hasFieldError = (
+    formikProps: FormikProps<FormikValues>,
+    fieldName: keyof FormikValues
+  ): boolean => getIsInvalid<FormikValues>(formikProps, fieldName, true);
+
+  const getFieldErrorMessage = (
```
VilleMiekkoja, line 58: I find the naming of this function little problematic. It is that it does not seem to return the FieldErrorMessage if there is no fieldError? One option would be to either move the check for _hasFieldError_ outside of this function, and then only call this function when we are sure it will return the FieldErrorMessage. Other option would be to name it differently: _getFieldErrorMessageIfExists_ or _getFieldErrorMessageOrUndefined_. Those names would give hint to user of the function that there is a chance it won't return the FieldErrorMessage

src/profile/components/basicData/BasicData.tsx
```diff
@@ -1,28 +1,172 @@
...
+    formikProps: FormikProps<FormikValues>,
+    fieldName: keyof FormikValues
+  ) => {
```
VilleMiekkoja: This seems to be missing typings for the return value?

src/profile/components/basicData/BasicData.tsx
```diff
@@ -1,28 +1,172 @@
...
+    if (!hasFieldError(formikProps, fieldName)) {
+      return undefined;
+    }
+    return getFieldError<FormikValues>(t, formikProps, fieldName, true);
+  };
+
+  const onAction: ActionListener = async (action, item, newValue) => {
+    clearMessage();
```
VilleMiekkoja, line 69: Would it be good idea to move the _useNotificationContent_ function call to line 67, so clearMessage used here would be defined closer to usage?

src/profile/components/basicData/BasicData.tsx
```diff
@@ -1,28 +1,172 @@
...
   });
+  const [isEditing, setEditing] = useState(false);
   const { t } = useTranslation();
+  const {
+    content,
+    setErrorMessage,
+    setSuccessMessage,
+    clearMessage,
+  } = useNotificationContent();
   const testId = basicDataType;
```
VilleMiekkoja: Im not sure if renaming the basicDataType to testId provides additional value? Could we apply the basicDataType directly instead?

src/profile/components/basicData/__tests__/BasicData.test.tsx
```diff
@@ -26,36 +35,157 @@ describe('<BasicData /> ', () => {
       </RenderChildrenWhenDataIsComplete>
     );
   };
+  const t = i18n.getFixedT('fi');
+  const editButtonSelector: ElementSelector = { id: 'basic-data-edit-button' };
+  const cancelButtonSelector: ElementSelector = {
```
VilleMiekkoja: cancelButtonSelector seems to be only used in the last test. Would it increase readability to move it inside the last test, as it seems not to be needed elsewhere yet?

src/profile/components/basicData/__tests__/BasicData.test.tsx
```diff
@@ -1,22 +1,31 @@
...
 } from '../../../../common/test/MockApolloClientProvider';
-import { BasicDataValue } from '../../../helpers/editData';
+import { basicDataType, BasicDataValue } from '../../../helpers/editData';
+import i18n from '../../../../common/test/testi18nInit';
 
 describe('<BasicData /> ', () => {
-  const renderTestSuite = (responses: MockedResponse[]) => {
+  const responses: MockedResponse[] = [];
+  const initialProfile = getMyProfile().myProfile as ProfileData;
+  const renderTestSuite = () => {
```
VilleMiekkoja, line 28: I would probably move this closer to usage, to line 92, above _initTests_ function. Same with initialProfile  object above at line 27.

src/profile/components/basicData/__tests__/BasicData.test.tsx
```diff
@@ -26,36 +35,157 @@ describe('<BasicData /> ', () => {
...
+      const { getTextOrInputValue, triggerAction } = await initTests();
+      await verifyValues(getTextOrInputValue, initialProfile);
+      // goto edit mode
+      await triggerAction(editButtonSelector);
+      await verifyValues(getTextOrInputValue, initialProfile, true);
+    });
+  });
+  it('sends new data and returns to view mode when saved', async () => {
+    // new values
+    const basicData = {
```
VilleMiekkoja: This _basicData_ initialisation seems to be made in the beginning of each test. I wonder if we could initialise it inside the _beforeEach_?

src/profile/components/editingNotifications/EditingNotifications.module.css
```diff
@@ -0,0 +1,10 @@
+
+.wrapper {
+  padding-bottom: var(--spacing-m);
+  /*Fixes notification rendering above language selection*/
+  z-index: 1;
+  position: relative;
+}
+.wrapper:focus {
+ outline:none;
```
VilleMiekkoja: Indentation and spacing seems little off. Think it would be a good idea to setup automated css lint script that would catch these. Maybe also prettier that would automatically format these correctly. Hauki applies stylelint library for this: https://www.npmjs.com/package/stylelint

src/profile/components/editingNotifications/EditingNotifications.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+  content: NotificationContent;
+  dataType: EditDataType;
+};
+
+function EditingNotifications({
+  content,
+  dataType,
+}: Props): React.ReactElement | null {
+  if (!content.text) {
```
VilleMiekkoja, line 17: According to clean code passing or returning null is questionable: https://enos.itcollege.ee/~jpoial/oop/naited/Clean%20Code.pdf#page=141

Because then you have to do these checks in the beginning of components and they don't render what you would expect them to render. Maybe one option is to check the content.text value in BasicData.tsx, and only render this EditingNotifications if it has a value 🤔

src/profile/components/editingNotifications/useNotificationContent.ts
```diff
@@ -0,0 +1,65 @@
...
+      return error
+        ? `${t('notification.saveError')} ${t(tryAgainKey)}`
+        : t('notification.saveSuccess');
+    }
+    return error
+      ? `${t('notification.genericError')} ${t(tryAgainKey)}`
+      : t('notification.genericSuccess');
+  };
+  const setSuccessMessage: ContentHandlers['setSuccessMessage'] = (
+    message,
```
VilleMiekkoja: Is this message used? In BasicData it feels like it is always set to an empty string. For example in line 41, it seems the callers always use action, so the result is never the message directly? I wonder if it would simplify the code if we removed that unused parameter?

src/profile/components/formButtons/FormButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+function FormButtons(props: Props): React.ReactElement {
+  const { t } = useTranslation();
+  const { handler, alignLeft, disabled, testId } = props;
+  return (
+    <div
+      className={classNames([
+        commonFormStyles.editButtons,
+        commonFormStyles.editButtons,
```
VilleMiekkoja: Hmm.. is this twice on purpose?

src/profile/components/saveIndicator/SaveIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type Props = {
+  action: Action | undefined;
+  testId: string;
+};
+
+const SaveIndicator = (props: Props): React.ReactElement | null => {
+  const { action, testId } = props;
+  const { t } = useTranslation();
+  if (!action) {
```
VilleMiekkoja, line 15: I would prefer this check to be in BasicData.tsx, but it can be just a preference thing. If you render the SaveIndicator component now, you can't be sure that it actually renders what it says, but it can return null also. I just think that the logic for whether we should show SaveIndicator or not belongs to a higher level than to the SaveIndicator component itself. Or then maybe one option could be to make the api of the SaveIndicator more clear, something like _showSaveIndicator_ prop that would indicate if it's rendered more easily. Now users of this component needs to dig in to the implementation of the component to see if it's actually rendered.

src/profile/components/saveIndicator/SaveIndicator.tsx
```diff
@@ -0,0 +1,31 @@
...
+    return null;
+  }
+  const translationKey =
+    action === 'remove' ? 'notification.removing' : 'notification.saving';
+  return (
+    <div
+      role="alert"
+      className={commonFormStyles.visuallyHidden}
```
VilleMiekkoja, line 23: Is the SaveIndicator hidden from screen?

NikoHelle: When form is submitted, buttons are disabled while saving. SaveIndicator is just for screen readers to indicate that submit happened and what kind of action (save/remove) is being done.

src/profile/components/formButtons/FormButtons.tsx
NikoHelle: No :) Will be fixed.

src/profile/components/editingNotifications/useNotificationContent.ts
NikoHelle: Good point, it was added for one situation which does not exist anymore. Will be removed

src/profile/components/editingNotifications/EditingNotifications.tsx
NikoHelle: I think it is normal in React world to render nothing (=return null), if component don't have correct data etc. to render its content. Developer can see it in return type definition.

This is a common component which will be used by also by address, phones, emails editors. It is easier and clearer to check content here than every time component is used. 

BasicData:
{ if !context.text && <EditingNotifications .../> }

Addresses:
{ if !context.text && <EditingNotifications .../> }

and so on...

And if there is a new prop (content.alert) that must be added to each implementation instead of in the component itself.

{ if !context.text && !content.alert && <EditingNotifications .../> }

src/profile/components/editingNotifications/EditingNotifications.module.css
NikoHelle: Ah, we have prettier. My IDE shows an error with the plugin, so it was not run.

src/profile/components/basicData/BasicData.tsx
NikoHelle: TS compiler / linter and Sonar would alert if needed. TS understands the return type is the return type of the  function called: "return getFieldError(....)"

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -102,5 +165,187 @@ describe('editData.ts ', () => {
...
+
+        // new value does not match old
+        expect(newValue).not.toMatchObject(oldValue);
+        // old target is not mutated
+        expect(target.value).toEqual(oldValue);
+        // new value matches value in editData
+        expect(newTarget.value).toEqual(newValue);
+        // ...and in formValues
+        expect(newFormValue).toMatchObject(newValue);
+        // new old old id and primary match
```
VilleMiekkoja: new id and primary match with old one?

src/profile/components/basicData/BasicData.tsx
NikoHelle: Isn't typescript return types enough to tell what function returns? :) In this case it is not written here but seen when you hover over the function. And where ever the function is called, you can see return types by hovering over the function call. This is editor specific of course, but arent' they visible in all normal editors?

I find it hard to put all return types to the function name itself.

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -102,5 +165,187 @@ describe('editData.ts ', () => {
         });
       }
     });
+    it(`updateEditData clones and updates data when dataType is ${dataType} and returns formValues`, () => {
```
VilleMiekkoja: Is the function being tested updateItemAndCreateSaveData? I wonder if updateEditData in this sentence should be replaced with updateItemAndCreateSaveData?

src/graphql/generatedTypes.ts
NikoHelle: Yes, there is codegen that fetches these monsters from graphQL server and UI cannot rename these. Typings.ts was added to make common "shortcuts".

src/common/test/myProfileMocking.ts
NikoHelle: They are grouped in UI as "Perustiedot". That's why using "Basic data". It might get additional fields that are not names. Language was in it too, but moved to "Lisätiedot".

src/common/test/myProfileMocking.ts
NikoHelle: Not needed, because TS can combine returned object and function's return type (ManipulationFunctions)

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -102,5 +165,187 @@ describe('editData.ts ', () => {
...
+        sources,
+        dataType
+      ) as EditData[];
+      expect(newList).toHaveLength(editDataList.length);
+      const updatedItem = newList[0];
+      // change saving so item will match updatedItem
+      expect(updatedItem).toEqual({ ...item, saving: undefined });
+      expect(updatedItem.saving).toEqual(undefined);
+    });
+    it(`Added new items are not lost and are updated from source'`, () => {
```
VilleMiekkoja: Should we skip this test for now or add todo or jira id? It might be confusing looking test results saying this test pass when it does not do anything yet?

src/common/test/MockApolloClientProvider.tsx
NikoHelle: GlobalWithFetchMock still exists, but you're right, documentation has changed regarding the setup. I'll try to remove  GlobalWithFetchMock and use enableMocks()

src/common/test/componentMocking.tsx
NikoHelle: The idea of this component is to start fetching and only once. If useEffect was used, "fetch" must be hook's dependency and I can't guarantee if the Apollo's fetch function never changes between renders. Change would re-run the useEffect and fetch.

src/common/test/MockApolloClientProvider.tsx
NikoHelle: Yes, it mocks all calls - but only when the test includes the MockApolloClientProvider.  And only Apollo graphQL-queries are fetched in the UI. It is true that other fetches are mocked too, but at the moment there aren't any and then MockApollo should not be used. Is the risk too big?

src/profile/helpers/editData.ts
```diff
@@ -80,17 +87,42 @@ export type EditData = {
...
+    targetItem: EditData,
+    newValue: EditDataValue
+  ) => Partial<FormValues>;
+  updateData: (newProfileRoot: ProfileRoot) => boolean;
+  updateAfterSavingError: (id: string) => boolean;
+  resetItem: (targetItem: EditData) => boolean;
 };
 
-function isMultiItemDataType(dataType: EditDataType): boolean {
+export function isMultiItemDataType(dataType: EditDataType): boolean {
```
VilleMiekkoja: Is this export necessary? Tried searching but did not find any usages outside of this file?

src/profile/helpers/editData.ts
```diff
@@ -80,17 +87,42 @@ export type EditData = {
...
+  return !!allItems.find(item => !!item.saving);
+}
+
+export function getNewItem(allItems: EditData[]): EditData | undefined {
+  return allItems.find(item => item.id === '');
+}
+
+export function hasNewItem(allItems: EditData[]): boolean {
+  return !!getNewItem(allItems);
+}
```
VilleMiekkoja: Did not find any usages of these functions outside of this file, so the exports are probably not needed?

src/profile/helpers/editData.ts
```diff
@@ -127,7 +159,7 @@ function create(
   };
```
VilleMiekkoja, line 159: Comment for line 155: According to my ide the line could be simplified to:

`primary: (profileData as MultiItemProfileNode).primary`

src/profile/helpers/editData.ts
```diff
@@ -105,15 +137,15 @@ function getValueProps(dataType: EditDataType): string[] {
...
+export function pickValue(
   profileDataItem: EditDataProfileSource,
   dataType: EditDataType
 ): EditDataValue {
   const pickProps = getValueProps(dataType);
   return _.pick(profileDataItem, pickProps) as EditDataValue;
 }
 
-function create(
+function createNewItem(
```
VilleMiekkoja, line 148: The function definition (line 148) is pretty far away from where it is used first time in this file: line 303. I wonder if it would be possible to move it closer to usage and if it would increase readability?

src/profile/helpers/editData.ts
```diff
@@ -127,7 +159,7 @@ function create(
   };
 }
 
-function pickSources(
+export function pickSources(
```
VilleMiekkoja, line 162: First usage of this function inside this file is pretty far away from the definition 162 -> 362. I wonder if it's possible to move it closer to usage to help readability? Sometimes can be hard if there's many functions that if you move one function closer, then some other function will be more far away 😅

src/common/test/MockApolloClientProvider.tsx
VilleMiekkoja: Probably not a big risk, just wondering if it could cause problems in the future where developer would not notice why is his fetch mocked if he tries testing code with other fetches.

src/common/test/myProfileMocking.ts
```diff
@@ -155,3 +207,189 @@ export const getVerifiedData = (
   },
   ...overrides,
 });
+
+function createNewProfileNode(dataType: EditDataType): MultiItemProfileNode {
```
VilleMiekkoja: There seems to be almost similar function defined in editData.ts line 227:

`export function createNewProfileNode<T extends MultiItemProfileNode>(...`

I wonder if we could apply it here some way?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
   }
 }
 
+export function cloneAndMutateItem(
```
VilleMiekkoja: Did not find any usages outside of this file, so maybe the export is unnecessary?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
...
+  data: EditData,
+  overrides?: Partial<EditData>
+): EditData {
+  return {
+    ...data,
+    ...overrides,
+  };
+}
+
+export function findItemIndex(
```
VilleMiekkoja: Did not find any usages outside of this file, so maybe the export is unnecessary?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
...
+  const itemId =
+    typeof idOrEditData === 'string' ? idOrEditData : idOrEditData.id;
+  return allItems.findIndex(item => itemId === item.id);
+}
+
+function findItem(allItems: EditData[], id: string): EditData | undefined {
+  return allItems.find(item => id === item.id);
+}
+
+export function updateItemAndCloneList(
```
VilleMiekkoja: Did not find any usages outside of this file, so maybe the export is unnecessary?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
...
+  const updatedItem = cloneAndMutateItem(item, {
+    value: newValue,
+    saving: saving || undefined,
+  });
+  const newList = _.cloneDeep(allItems);
+  newList[index] = updatedItem;
+  return newList;
+}
+
+export function createNewProfileNode<T extends MultiItemProfileNode>(
```
VilleMiekkoja, line 227: Did not find any usages outside of this file, so maybe the export is unnecessary? Though there's one similar looking function at myProfileMocking.ts line 211, maybe these could be combined?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
...
+  dataType: EditDataType,
+  overrides?: Partial<T>
+): T {
+  return {
+    ...(formConstants.EMPTY_VALUES[dataType] as T),
+    ...overrides,
+  };
+}
+
+export function createFormValues(
```
VilleMiekkoja: Did not find any usages outside of this file, so maybe the export is unnecessary?

src/profile/helpers/editData.ts
```diff
@@ -150,17 +182,254 @@ function pickSources(
...
+  if (newItem) {
+    newList.push(newItem);
+  }
+  if (!dataHasUpdated && allItems.length !== profileDataItems.length) {
+    dataHasUpdated = true;
+  }
+  return dataHasUpdated ? newList : null;
+}
+
+export function createBackups(): Backups {
```
VilleMiekkoja: Did not find any usages outside of this file, so maybe the export is unnecessary?

src/profile/helpers/updateMutationVariables.ts
```diff
@@ -194,33 +199,48 @@ function formMutationArrays<T extends AddressNode | EmailNode | PhoneNode>(
...
+        ...emailData,
+        ...addressData,
+      },
     },
-  },
-});
+  };
+};
 
 export { formMutationArrays, updateMutationVariables };
```
VilleMiekkoja, line 246: It seems in some files the exports are inlined, and in some files like this they are at the end of the file. I wonder if it would be good idea to use one style of exporting mainly? Inlined exports seems to be more common maybe?

src/profile/helpers/updateMutationVariables.ts
```diff
@@ -1,4 +1,4 @@
-import { isEqual } from 'lodash';
+import _, { isEqual } from 'lodash';
```
VilleMiekkoja: Should we import the pick separately like in the case of isEqual? Or just '_' to handle both?

src/profile/helpers/updateMutationVariables.ts
VilleMiekkoja: Also did not find usages for _formMutationArrays_. Maybe exporting it is unnecessary?

src/profile/hooks/useProfileDataEditor.ts
```diff
@@ -1,34 +1,109 @@
...
 } from '../helpers/editData';
+import { UpdateResult, useProfileMutations } from './useProfileMutations';
 
 export type ProfileDataEditorReturnType = {
   editDataList: EditData[];
+  save: (item: EditData, newValue: EditDataValue) => Promise<UpdateResult>;
+  reset: (item: EditData) => void;
 };
 
+export type Action = SaveType | 'edit' | 'cancel' | 'save' | 'add';
```
VilleMiekkoja, line 20: What's the difference between SaveType and 'save'? Sounds semantically similar.

src/profile/hooks/useProfileDataEditor.ts
```diff
@@ -1,34 +1,109 @@
...
-  const { getEditData } = useMemo(
+  const {
+    getEditData,
+    updateItemAndCreateSaveData,
+    updateData,
+    updateAfterSavingError,
+    resetItem,
+  } = useMemo(
     () => createEditorForDataType(profileData, dataType),
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
VilleMiekkoja, line 63: Why do we not want to recompute if profileData or dataType has changed?

src/profile/hooks/useProfileDataEditor.ts
```diff
@@ -1,34 +1,109 @@
...
     []
   );
 
+  const [, setUpdateTime] = useState<number>(Date.now());
+  const triggerUpdate = () => {
+    setUpdateTime(Date.now());
```
VilleMiekkoja, line 69: Not sure what is the purpose of this triggerUpdate? Is it to just manually trigger UI update by applying useState? It kind of feels like a hacky thing to do. I wonder if there's any alternatives 🤔

src/profile/hooks/useProfileMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+import {
+  FormValues,
+  EditDataType,
+  basicDataType,
+  additionalInformationType,
+} from '../helpers/editData';
+import { updateMutationVariables } from '../helpers/updateMutationVariables';
+import { ProfileContext, ProfileContextData } from '../context/ProfileContext';
+
+const UPDATE_PROFILE = loader('../../profile/graphql/UpdateMyProfile.graphql');
```
VilleMiekkoja, line 27: Is there a reason why these follow different naming conventions than regular const variables?

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Yes, when data is updated in some way that does not make hook re-render (like pressing cancel) the re-render must be triggered manually. Same kind of easy re-render is used for example in MobX:

https://github.com/mobxjs/mobx-react-lite/blob/master/src/utils/utils.ts

src/profile/hooks/useProfileMutations.ts
NikoHelle: Only reason is that the old Profile UI had them like that the code is moved here.

src/profile/hooks/useProfileMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+  profileData?: ProfileRoot;
+  update: UpdateProfileFunction;
+};
+
+type UpdatableProfileDataKeys = keyof Omit<
+  UpdateProfileData,
+  '__typename' | 'id'
+>;
+
+export type ValuesOfUpdateProfileData = UpdateProfileData[UpdatableProfileDataKeys];
```
VilleMiekkoja: Did not find any usages of this type outside of this file, so maybe the export is unnecessary?

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Don't want to override current data. It may include new data items (in case of addresses, emails, phones) or unsaved data that are not in profileData. Update is done separately in useEffect.

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: SaveType in EditData tells what property is being saved: 'values' or 'set-primary'.  'save' is just an general action.

src/profile/helpers/updateMutationVariables.ts
NikoHelle: Auto-import did this. _.isEqual is better.

src/profile/hooks/useProfileMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+  const update: UpdateProfileFunction = (formValues, profile) => {
+    const currentProfile = profile || profileData;
+    if (!currentProfile) {
+      return Promise.reject({ error: 'No profile data' });
+    }
+    const variables = updateMutationVariables(formValues, currentProfile);
+    return updateProfile({ variables });
+  };
+
+  return { update, error, loading, profileData };
```
VilleMiekkoja: Did not find usages for error and loading, maybe there's no need to return those?

src/profile/helpers/updateMutationVariables.ts
NikoHelle: This is an old file made by someone else. Haven't followed their preference of exporting. I think exporting where the function actually is, is much better. Won't find all occurences and change them to match that 😄

src/profile/hooks/useProfileMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+  profileData: ProfileRoot | null | undefined,
+  updateProfileData: ProfileContextData['updateProfileData']
+) {
+  const updatedProfile: UpdateProfileData | null | undefined =
+    result.data?.updateMyProfile?.profile;
+  const identity =
+    profileData &&
+    cache.identify((profileData.myProfile as unknown) as Reference);
+  if (!updatedProfile || !identity) {
+    throw new Error('Unable to update cache2');
```
VilleMiekkoja: cache2? This might be that Im beginner related to graphql, but is there cache1 somewhere?

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -106,112 +108,251 @@ const newAddress = {
...
-  expect(variables.input.profile.updateAddresses).toEqual([
-    getAddressAsComparisonObject(updatedSecondaryAddress),
-  ]);
-  expect(variables.input.profile.updateEmails).toEqual([
-    getEmailAsComparisonObject(updatedSecondaryEmail),
-  ]);
-  expect(variables.input.profile.updatePhones).toEqual([
-    getPhoneAsComparisonObject(updatedSecondaryPhone),
-  ]);
+  test('Variables in update data match formValues even if not changed', () => {
```
VilleMiekkoja: There seems to be 2 different conventions for this test function name: for example editData.test.ts applies function _it_. I wonder if it would be a good idea to only use one of them?

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -106,112 +108,251 @@ const newAddress = {
...
+    const variables: UpdateMyProfileVariables = updateMutationVariables(
+      formValues,
+      myProfile
+    );
```
VilleMiekkoja, line 152: This variables initialisation seems to be done in multiple tests (in 7 total). Not sure if it could be initialised in beforeEach, and then in tests that do not wish to use it, they would reinitialise it differently. Though sometimes repetition actually improves readability it feels, so not sure 🤔

src/profile/components/basicData/BasicData.tsx
VilleMiekkoja: Ide and hovering definitely helps. But I find it easier to read if I can determine what the function does (returns in this case) from the name. Hovering over is an extra step that slows down reading, and should be only needed if I wish to know more details. I agree that a function with many return types in its name feels bad. I think in general it would be good for a function to only return one types of things. The more types the function returns, the more complex it is, and the more harder it is for the reader of the code. I would probably prefer moving the check for hasFieldError outside of this function for that reason.  But yeah obviously this is very opinion dependant when it comes to readability and what is enough.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -106,112 +108,251 @@ const newAddress = {
...
-      emails: [primaryEmail, newEmail],
-      phones: [primaryPhone, newPhone],
-    },
-    myProfile
-  );
-
-  expect(variables.input.profile.updateAddresses).toEqual([]);
-  expect(variables.input.profile.updateEmails).toEqual([]);
-  expect(variables.input.profile.updatePhones).toEqual([]);
+describe('Update variable for language is formed correctly', () => {
```
VilleMiekkoja, line 185: Minor notice about language: It seems in some of the tests the first letter is capitalised and in some tests it's not. Not a big thing though

src/profile/helpers/__tests__/updateMutationVariables.test.ts
```diff
@@ -106,112 +108,251 @@ const newAddress = {
...
+    expect(variables.input.profile.language).toBeDefined();
+    expect(variables.input.profile.updateAddresses).toBeDefined();
+    expect(variables.input.profile.updateEmails).toBeDefined();
+    expect(variables.input.profile.updatePhones).toBeDefined();
+  });
 
-  expect(variables.input.profile.removeAddresses).toBeFalsy();
-  expect(variables.input.profile.removeEmails).toBeFalsy();
-  expect(variables.input.profile.removePhones).toBeFalsy();
+  test('but empty update is also handled', () => {
```
VilleMiekkoja: This feels like a side sentence to another sentence? It might be a good idea to make each test and its name independent of other test names for clarity. If the other test is then removed or modified, this one might not make as much sense anymore. Maybe just remove the first word _but_?

VilleMiekkoja:
Tested it locally and seems to be working nice! Coding this has required lot of effort, well done! 🎖️

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Yes, "test"-funcs are used in the old project files. I kept the same func calls inside each file. I'll rename them.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Yes, sometimes it's a typo, usually the case is:

describe("This describes the purpose of the test group ')
it('and this is in lower case becasue it continues the sentence started in describe()")

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Yeah, there are 7 of those code blocks and in between 5 that does different call. That's why the call is not in beforeEach. Those 7 cannot be logically grouped under one describe(). I thought about using a helper function, but then the readability came to mind. Would it be harder to follow the test if some tests had different initialisation function than others?

Instead of 

```
const variables: UpdateMyProfileVariables = updateMutationVariables(
      formValues,
      myProfile
    );
```
it was something like

`createCommonUpdateVariables();`

Or create one helper function for all tests - with optional formValues:

`createUpdateVariables(formValues?:FormValues);`

Better or worse? :)

src/profile/hooks/useProfileMutations.ts
NikoHelle: That's a relic from some debugging session. :)

src/profile/hooks/useProfileMutations.ts
NikoHelle: Except in useProfileMutations.test.ts. UI does not use them because Promise.reject() is catched instead. But in tests error and loading are directly tested. Should they be kept there?

src/common/test/myProfileMocking.ts
NikoHelle: Good catch!

src/profile/helpers/editData.ts
NikoHelle: Yes, I merged them.

src/profile/helpers/editData.ts
NikoHelle: Referring to the Team's Slack discussion, ordering functions and keeping them in logical order for everyone is a tedious job and hard to keep the ordering in the long run.

We should write the code so functions are in clear, logical order and try to keep them so, but it is not required and enforced.

src/profile/helpers/editData.ts
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/profile/helpers/editData.ts
NikoHelle: Correct, they (like other unused exports) were exported for testing, but they are tested via functions that use them.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Ok, I left it there as a placeholder. Adding new items is in next PR. I'll delete this.

src/profile/components/saveIndicator/SaveIndicator.tsx
NikoHelle: Same comment than above: components that are used by multiple components, should make these check internally in once place instead of in all components that use it and it's ok to return null.

https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r617231848

src/profile/helpers/__tests__/updateMutationVariables.test.ts
VilleMiekkoja: One option could be to make 1 more master describe on top of those all, and inside that there would be beforeEach. Not sure if it would make things more complicated though. Repetition of those initialisations like it is now might be best solution for readability

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: I thought it was cleaner to group selectors in one place. I'll change this.

src/profile/components/basicData/BasicData.tsx
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/profile/hooks/useProfileMutations.ts
VilleMiekkoja: Are there plans on using them in the UI in future? If not, it might simplify things to delete them if they are not needed.

src/profile/components/basicData/BasicData.tsx
NikoHelle: Two reasons why hasFieldError is in getFieldErrorMessage

1: it is used in an element-prop:
helperText={getFieldErrorMessage(formikProps, 'nickname')}

in this case it should be changed to:
helperText={hasFieldError(formikProps, 'nickname') && getFieldErrorMessage(formikProps, 'nickname')}

that's looks terrible IMHO. And that in 3 places in this file and later on in address, postal code, city, email and phone. 

2: it has always been there. It this file it looks like it was created here, but it's a copy from old code

NikoHelle: Should I copy the return type of getFieldError to getFieldErrorMessage? Copying code for faster readability is not very convenient. When it getFieldError changes, it must be changed everywhere. Later on also in addresses, emails and phones will use it.

I can make a common type for them, but seems a bit excessive. You cannot read it in IDE either.

src/profile/components/basicData/BasicData.tsx
NikoHelle: No, because React hooks ("useNotificationContent") cannot be run conditionally.

src/common/test/myProfileMocking.ts
NikoHelle: "clonedProfileWithManipulationFunctions" sounds like a return type?

Renamed to cloneProfileAndProvideManipulationFunctions (which is a long monster :D ) IMHO the actual return type describes the function already quite well. But that is only seen via hover in the IDE.

src/common/test/myProfileMocking.ts
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/common/test/myProfileMocking.ts
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/profile/components/basicData/BasicData.tsx
VilleMiekkoja: I meant that useProfileDataEditor function call would be above this condition, at line 44

src/profile/components/basicData/BasicData.tsx
VilleMiekkoja: To my eye this:

```helperText={hasFieldError(formikProps, 'nickname') && getFieldErrorMessage(formikProps, 'nickname')}```

doesn't look that bad because you then know the condition when there will be a helperText. 😅 But I guess this is about what reader of the code is accustomed to. I've probably just used to read code where logic is shown in higher abstraction level. I think the current solution is good enough, we can go with that. 👍

src/common/cssHelpers/form.module.css
NikoHelle: All old profile css files are named *.module.css. Probably because name is required so React identifies them as css modules. 

https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

src/common/test/componentMocking.tsx
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/common/test/componentMocking.tsx
NikoHelle: In normal use, why would you use multiple selectors at the same time? Like id + testid, label + text or id  + text. QuerySelector with label/text/value would make sense, but others not. When selector with value is needed there is waitForElementAndValue.
Selector with label/text is not currently needed.

I added this so there is only one function to get an element.  getElement is also used in tests outside of this file. If it did not exists, all funcs from testing-library should be passed through to tests
getByLabelText(label);
getByTestId(testId);
getByText(text);
getByDisplayValue(valueSelector);
(+ create new ones for id and querySelector )

GetElement is easier to use that than finding each time the correct one from testing-library.

src/common/test/componentMocking.tsx
NikoHelle: Renamed to renderComponentWithMocksAndContexts. The return type in TS-files should be enough to tell what a function returns instead of using renderComponentWithMocksAndContextsAndReturnTestTools 😄 

All tested components need Router, mocked GraphQL, Profile Context, and Toast Context. And ability to fetch data, find elements, interact with DOM and waitFor something to be complete. This test helper returns context and tools for those. 

Plan B would be to render components in one func and pass result of that to another function(s) to do the rest. In each test file.

src/common/test/componentMocking.tsx
NikoHelle: Referring to another re-ordering comment:
https://github.com/City-of-Helsinki/open-city-profile-ui/pull/154#discussion_r618158453

src/common/test/MockApolloClientProvider.tsx
NikoHelle: I'll change this to catch only /graphql/ paths.

src/profile/hooks/useProfileMutations.ts
NikoHelle: Ok, I removed them and changed the test just to  check error wont alter profile data

src/common/test/MockApolloClientProvider.tsx
NikoHelle: Done.

src/common/test/componentMocking.tsx
NikoHelle: Done.

src/common/test/MockApolloClientProvider.tsx
NikoHelle: Global stuff removed!

src/common/test/myProfileMocking.ts
NikoHelle: Done.

src/profile/components/basicData/BasicData.tsx
NikoHelle: Ah... Done.

src/profile/components/basicData/BasicData.tsx
NikoHelle: I left this as it is.

src/profile/components/basicData/BasicData.tsx
NikoHelle: Done

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: Done.

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: Done.

src/profile/components/editingNotifications/EditingNotifications.module.css
NikoHelle: Ran all css (old and new) through prettier. All changed!

src/profile/components/editingNotifications/useNotificationContent.ts
NikoHelle: Done.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Typo fixed.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Fixed typo

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Removed

src/profile/helpers/editData.ts
NikoHelle: Removed

src/profile/helpers/editData.ts
NikoHelle: Removed

src/profile/helpers/editData.ts
NikoHelle: Changed. Affected tests and fixed them.

src/common/test/myProfileMocking.ts
NikoHelle: Done.

src/profile/helpers/editData.ts
NikoHelle: Removed.

src/profile/helpers/editData.ts
NikoHelle: Removed.

src/profile/helpers/editData.ts
NikoHelle: Removed.

src/profile/helpers/editData.ts
NikoHelle: Removed.

src/profile/helpers/editData.ts
NikoHelle: Removed.

src/profile/helpers/updateMutationVariables.ts
NikoHelle: Done.

src/profile/hooks/useProfileMutations.ts
NikoHelle: Removed.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Done.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Left as it is.

src/profile/helpers/__tests__/updateMutationVariables.test.ts
NikoHelle: Fixed.

src/profile/components/formButtons/FormButtons.tsx
NikoHelle: Done.

NikoHelle:
All changes should be done now. Didn't make function re-orderings.

VilleMiekkoja (approved):
Well done! 👍

PR #156, opened 2021-04-26
HP-318 step 6:  render and edit emails, phones and addresses

NikoHelle:
Emails, phones and addresses are now visible in the UI and can be edited. New items can be added and removed by clicking cancel button. Old, saved data cannot be removed yet.


NikoHelle:
Rebase should be done first.

PR #157, opened 2021-04-26
HP-318 step 6: add and edit emails, phones and addresses

NikoHelle:
Added components for viewing and editing emails, phones and addresses. 

Old data can be saved and new added and saved, but "remove"-button is not added for deleting items.

If there is a need to remove data, it can be done in https://profiili.dev.hel.ninja/

### DIFF
--- a/src/common/test/componentMocking.tsx
+++ b/src/common/test/componentMocking.tsx
@@ -48,7 +48,7 @@ export type RenderHookResultsChildren = {
 
 export type WaitForElementAndValueProps = {
   selector: ElementSelector;
-  value?: string;
+  value: string;
 };
 
 export type TestTools = RenderResult & {
@@ -60,11 +60,12 @@ export type TestTools = RenderResult & {
     selector: ElementSelector
   ) => Promise<string | undefined>;
   fetch: () => Promise<void>;
-  clickElement: (selector: ElementSelector) => Promise<void>;
+  clickElement: (selector: ElementSelector) => Promise<HTMLElement | null>;
   submit: (props?: {
     waitForOnSaveNotification?: WaitForElementAndValueProps;
     waitForAfterSaveNotification?: WaitForElementAndValueProps;
     skipDataCheck?: boolean;
+    optionalSubmitButtonSelector?: ElementSelector;
   }) => Promise<void>;
   isDisabled: (element: HTMLElement | Element | null) => boolean;
   setInputValue: ({
@@ -86,6 +87,23 @@ export const cleanComponentMocks = (): void => {
 };
 
 export const emptyResponseProvider: ResponseProvider = () => ({});
+export const submitButtonSelector: ElementSelector = {
+  querySelector: 'button[type="submit"]',
+};
+
+export const waitForElementAttributeValue = async (
+  elementGetter: () => HTMLElement | Element | null,
+  attribute: string,
+  value: string
+): Promise<void> => {
+  await waitFor(async () => {
+    const element = elementGetter();
+    const attributeValue = element && element.getAttribute(attribute);
+    if (attributeValue !== value) {
+      throw new Error('Attribute value mismatch');
+    }
+  });
+};
 
 export const renderComponentWithMocksAndContexts = async (
   responseProvider: ResponseProvider,
@@ -199,7 +217,7 @@ export const renderComponentWithMocksAndContexts = async (
   const clickElement: TestTools['clickElement'] = async selector => {
     const button = getElement(selector);
     fireEvent.click(button as Element);
-    return Promise.resolve();
+    return Promise.resolve(button);
   };
 
   const isDisabled: TestTools['isDisabled'] = element =>
@@ -209,17 +227,17 @@ export const renderComponentWithMocksAndContexts = async (
     waitForOnSaveNotification,
     waitForAfterSaveNotification,
     skipDataCheck,
+    optionalSubmitButtonSelector,
   } = {}) => {
     const previousDataChangeTime = getLastTime('dataUpdateTime');
-    const submitButton = renderResult.container.querySelectorAll(
-      'button[type="submit"]'
+    const submitButton = await clickElement(
+      optionalSubmitButtonSelector || submitButtonSelector
     );
-    fireEvent.click(submitButton[0]);
     if (waitForOnSaveNotification) {
       await waitForElementAndValue(waitForOnSaveNotification);
     }
     await waitFor(() => {
-      if (!isDisabled(submitButton[0])) {
+      if (!isDisabled(submitButton)) {
         throw new Error('NOT DISABLED');
       }
     });
@@ -231,25 +249,30 @@ export const renderComponentWithMocksAndContexts = async (
     }
   };
 
+  const waitForElementAndValue: TestTools['waitForElementAndValue'] = async props => {
+    const { selector, value } = props;
+    return waitFor(async () => {
+      const elementValue = await getTextOrInputValue(selector);
+      if (value === '' && elementValue === '') {
+        return;
+      }
+      if (!elementValue || !elementValue.includes(value)) {
+        throw new Error('element value does not match given value');
+      }
+    });
+  };
+
   const setInputValue: TestTools['setInputValue'] = async props => {
     const { newValue, target, selector } = props;
     const input = target || getElement(selector);
     fireEvent.change(input as HTMLElement, { target: { value: newValue } });
 
     return waitFor(
-      () => {
-        getElement({ valueSelector: newValue });
+      async () => {
+        await waitForElementAndValue({ selector, value: newValue });
       },
       { timeout: 60 }
     );
-  };
-
-  const waitForElementAndValue: TestTools['waitForElementAndValue'] = async props => {
-    const { selector, value } = props;
-    const elementValue = await getTextOrInputValue(selector);
-    if (value && (!elementValue || !elementValue.includes(value))) {
-      throw new Error('element value does not match given value');
-    }
   };
 
   return Promise.resolve({
--- a/src/profile/components/basicData/__tests__/BasicData.test.tsx
+++ b/src/profile/components/basicData/__tests__/BasicData.test.tsx
@@ -12,6 +12,8 @@ import {
   cleanComponentMocks,
   WaitForElementAndValueProps,
   ElementSelector,
+  submitButtonSelector,
+  waitForElementAttributeValue,
 } from '../../../../common/test/componentMocking';
 import { ProfileData } from '../../../../graphql/typings';
 import BasicData from '../BasicData';
@@ -21,6 +23,7 @@ import {
 } from '../../../../common/test/MockApolloClientProvider';
 import { basicDataType, BasicDataValue } from '../../../helpers/editData';
 import i18n from '../../../../common/test/testi18nInit';
+import { getFormFields } from '../../../helpers/formProperties';
 
 describe('<BasicData /> ', () => {
   const responses: MockedResponse[] = [];
@@ -179,6 +182,76 @@ describe('<BasicData /> ', () => {
       });
       // values are reset to previous values
       await verifyValues(getTextOrInputValue, initialProfile);
+    });
+  });
+  it('invalid values are indicated and setting a valid value removes error', async () => {
+    await act(async () => {
+      const { clickElement, setInputValue, getElement } = await initTests();
+      const formFields = getFormFields(basicDataType);
+      await clickElement(editButtonSelector);
+      const testRuns = [
+        {
+          validData: basicData,
+          invalidData: { ...basicData, firstName: '' },
+          elementSelector: { id: 'basic-data-firstName' },
+          errorSelector: {
+            id: 'basic-data-firstName-helper',
+          },
+        },
+        {
+          validData: basicData,
+          invalidData: { ...basicData, lastName: '' },
+          elementSelector: { id: 'basic-data-lastName' },
+          errorSelector: {
+            id: 'basic-data-lastName-helper',
+          },
+        },
+        {
+          validData: basicData,
+          invalidData: {
+            ...basicData,
+            nickname: String('a').repeat(
+              (formFields.nickname.max as number) + 1
+            ),
+          },
+          elementSelector: { id: 'basic-data-nickname' },
+          errorSelector: {
+            id: 'basic-data-nickname-helper',
+          },
+        },
+      ];
+
+      // cannot use forEach with async/await
+      for (const runProps of testRuns) {
+        const {
+          validData,
+          invalidData,
+          elementSelector,
+          errorSelector,
+        } = runProps;
+        const elementGetter = () => getElement(elementSelector);
+        const errorElementGetter = () => getElement(errorSelector);
+
+        // set invalid values
+        await setValues(setInputValue, invalidData);
+        // submit also validates the form
+        await clickElement(submitButtonSelector);
+        await waitForElementAttributeValue(
+          elementGetter,
+          'aria-invalid',
+          'true'
+        );
+        // getElement throws if element is not found
+        expect(() => errorElementGetter).not.toThrow();
+        // set valid value
+        await setValues(setInputValue, validData);
+        await waitForElementAttributeValue(
+          elementGetter,
+          'aria-invalid',
+          'false'
+        );
+        expect(errorElementGetter).toThrow();
+      }
     });
   });
 });
--- a/src/profile/components/basicData/__tests__/BasicData.test.tsx
+++ b/src/profile/components/basicData/__tests__/BasicData.test.tsx
@@ -8,13 +8,12 @@ import {
 import {
   renderComponentWithMocksAndContexts,
   TestTools,
-  RenderChildrenWhenDataIsComplete,
   cleanComponentMocks,
   WaitForElementAndValueProps,
   ElementSelector,
   submitButtonSelector,
   waitForElementAttributeValue,
-} from '../../../../common/test/componentMocking';
+} from '../../../../common/test/testingLibraryTools';
 import { ProfileData } from '../../../../graphql/typings';
 import BasicData from '../BasicData';
 import {
@@ -24,6 +23,7 @@ import {
 import { basicDataType, BasicDataValue } from '../../../helpers/editData';
 import i18n from '../../../../common/test/testi18nInit';
 import { getFormFields } from '../../../helpers/formProperties';
+import RenderChildrenWhenDataIsComplete from '../../../../common/test/RenderChildrenWhenDataIsComplete';
 
 describe('<BasicData /> ', () => {
   const responses: MockedResponse[] = [];
--- a/src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
+++ b/src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
@@ -1,0 +1,199 @@
+import { TextInput } from 'hds-react';
+import React from 'react';
+import { Field, Formik, FormikProps, Form } from 'formik';
+import { useTranslation } from 'react-i18next';
+import countries from 'i18n-iso-countries';
+import classNames from 'classnames';
+
+import commonFormStyles from '../../../common/cssHelpers/form.module.css';
+import { AddressValue } from '../../helpers/editData';
+import { createFormFieldHelpers } from '../../helpers/formik';
+import { addressSchema } from '../../../common/schemas/schemas';
+import FormButtons from '../formButtons/FormButtons';
+import EditButtons from '../editButtons/EditButtons';
+import FormikDropdown, {
+  HdsOptionType,
+} from '../../../common/formikDropdown/FormikDropdown';
+import getLanguageCode from '../../../common/helpers/getLanguageCode';
+import LabeledValue from '../../../common/labeledValue/LabeledValue';
+import getCountry from '../../helpers/getCountry';
+import { getFormFields } from '../../helpers/formProperties';
+import SaveIndicator from '../saveIndicator/SaveIndicator';
+import { useCommonEditHandling } from '../../hooks/useCommonEditHandling';
+import { RowItemProps } from '../multiItemEditor/MultiItemEditor';
+
+type FormikValues = AddressValue;
+
+function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
+  const { data, testId, dataType } = props;
+  const value = data.value as AddressValue;
+  const { address, city, postalCode, countryCode } = value;
+  const { t, i18n } = useTranslation();
+  const lang = i18n.languages[0];
+  const applicationLanguage = getLanguageCode(i18n.languages[0]);
+  const countryList = countries.getNames(applicationLanguage);
+  const countryOptions = Object.keys(countryList)
+    .map(key => ({
+      value: key,
+      label: countryList[key],
+    }))
+    .sort((a, b) => {
+      if (a.label < b.label) {
+        return -1;
+      }
+      if (a.label > b.label) {
+        return 1;
+      }
+      return 0;
+    });
+  const formFields = getFormFields(dataType);
+  const {
+    isNew,
+    isEditing,
+    actionHandler,
+    currentAction,
+  } = useCommonEditHandling(props);
+  const { hasFieldError, getFieldErrorMessage } = createFormFieldHelpers<
+    FormikValues
+  >(t, isNew);
+
+  const { primary } = data;
+
+  if (isEditing) {
+    return (
+      <Formik
+        initialValues={{
+          address,
+          city,
+          postalCode,
+          countryCode,
+        }}
+        onSubmit={async values => {
+          await actionHandler('save', values);
+        }}
+        validationSchema={addressSchema}
+      >
+        {(formikProps: FormikProps<FormikValues>) => (
+          <Form className={commonFormStyles.multiItemForm}>
+            <h4 className={commonFormStyles.sectionTitle}>
+              {primary
+                ? t('profileInformation.primaryAddress')
+                : t('profileInformation.address')}
+            </h4>
+            <div className={commonFormStyles.multiItemWrapper}>
+              <Field
+                name="address"
+                id={`${testId}-address`}
+                maxLength={formFields.address.max as number}
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'address')}
+                aria-invalid={hasFieldError(formikProps, 'address')}
+                helperText={getFieldErrorMessage(formikProps, 'address')}
+                labelText={t(formFields.address.translationKey)}
+                autoFocus
+                aria-labelledby={`${dataType}-address-helper`}
+              />
+              <Field
+                name="postalCode"
+                id={`${testId}-postalCode`}
+                maxLength={formFields.postalCode.max as number}
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'postalCode')}
+                aria-invalid={hasFieldError(formikProps, 'postalCode')}
+                helperText={getFieldErrorMessage(formikProps, 'postalCode')}
+                labelText={t(formFields.postalCode.translationKey)}
+                aria-labelledby={`${dataType}-postalCode-helper`}
+              />
+              <Field
+                name="city"
+                id={`${testId}-city`}
+                maxLength={formFields.city.max as number}
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'city')}
+                aria-invalid={hasFieldError(formikProps, 'city')}
+                helperText={getFieldErrorMessage(formikProps, 'city')}
+                labelText={t(formFields.city.translationKey)}
+                aria-labelledby={`${dataType}-city-helper`}
+              />
+              <div className={commonFormStyles.formField}>
+                <FormikDropdown
+                  name="countryCode"
+                  id={`${testId}-countryCode`}
+                  options={countryOptions}
+                  label={t(formFields.country.translationKey)}
+                  default={countryCode}
+                  onChange={option =>
+                    formikProps.setFieldValue(
+                      'countryCode',
+                      (option as HdsOptionType).value
+                    )
+                  }
+                />
+              </div>
+            </div>
+            <FormButtons
+              handler={actionHandler}
+              disabled={!!currentAction}
+              testId={testId}
+              alignLeft
+            />
+            <SaveIndicator action={currentAction} testId={testId} />
+          </Form>
+        )}
+      </Formik>
+    );
+  }
+  return (
+    <div
+      className={classNames([
+        commonFormStyles.contentWrapper,
+        commonFormStyles.multiItemContentWrapper,
+      ])}
+    >
+      <h4 className={commonFormStyles.sectionTitle}>
+        {primary
+          ? t('profileInformation.primaryAddress')
+          : t('profileInformation.address')}
+      </h4>
+      <div className={commonFormStyles.multiItemWrapper}>
+        <LabeledValue
+          label={t(formFields.address.translationKey)}
+          value={value.address}
+          testId={`${testId}-address`}
+        />
+        <LabeledValue
+          label={t(formFields.postalCode.translationKey)}
+          value={value.postalCode}
+          testId={`${testId}-postalCode`}
+        />
+        <LabeledValue
+          label={t(formFields.city.translationKey)}
+          value={value.city}
+          testId={`${testId}-city`}
+        />
+        <LabeledValue
+          label={t(formFields.country.translationKey)}
+          value={getCountry(value.countryCode, lang)}
+          testId={`${testId}-countryCode`}
+        />
+      </div>
+      <div className={commonFormStyles.actionsWrapper}>
+        <EditButtons
+          handler={actionHandler}
+          actions={{
+            removable: false,
+            primary,
+            setPrimary: false,
+          }}
+          buttonClassNames={commonFormStyles.actionsWrapperButton}
+          editButtonId={`${testId}-edit-button`}
+          disabled={!!currentAction}
+          testId={testId}
+        />
+        <SaveIndicator action={currentAction} testId={testId} />
+      </div>
+    </div>
+  );
+}
+
+export default MultiItemAddressRow;
--- a/src/profile/components/multiItemEditor/MultiItemEditor.tsx
+++ b/src/profile/components/multiItemEditor/MultiItemEditor.tsx
@@ -1,0 +1,173 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import classNames from 'classnames';
+import to from 'await-to-js';
+import { Button, IconPlusCircle } from 'hds-react';
+
+import ProfileSection from '../../../common/profileSection/ProfileSection';
+import MultiItemRow from '../multiItemRow/MultiItemRow';
+import MultiItemAddressRow from '../multiItemAddressRow/MultiItemAddressRow';
+import commonFormStyles from '../../../common/cssHelpers/form.module.css';
+import useNotificationContent from '../editingNotifications/useNotificationContent';
+import EditingNotifications from '../editingNotifications/EditingNotifications';
+import {
+  useProfileDataEditor,
+  ActionListener,
+} from '../../hooks/useProfileDataEditor';
+import {
+  EditData,
+  EditDataType,
+  EditDataValue,
+  isNewItem,
+} from '../../helpers/editData';
+
+type Props = {
+  dataType: EditDataType;
+};
+
+export type RowItemProps = {
+  data: EditData;
+  onAction: ActionListener;
+  testId: string;
+  dataType: EditDataType;
+};
+
+function MultiItemEditor({ dataType }: Props): React.ReactElement | null {
+  const {
+    editDataList,
+    save,
+    reset,
+    add,
+    hasNew,
+    remove,
+  } = useProfileDataEditor({
+    dataType,
+  });
+  const { t } = useTranslation();
+  const {
+    content,
+    setErrorMessage,
+    setSuccessMessage,
+    clearMessage,
+  } = useNotificationContent();
+
+  const hasAddressList = dataType === 'addresses';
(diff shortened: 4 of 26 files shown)


src/common/test/componentMocking.tsx
```diff
@@ -60,11 +60,12 @@ export type TestTools = RenderResult & {
     selector: ElementSelector
```
VilleMiekkoja: File comment: This file is quite large. It seems to contain test related helper functions and types? I just wonder if it's doing too much, as the name componentMocking feels like it does not match for all what the file exposes. For example functions getElementAttribute and waitForElementAttributeValue. They feel like helper functions for testing, not specifically about mocking components? Not sure about how it could be improved though, just an observation 😅

src/common/test/componentMocking.tsx
```diff
@@ -86,6 +87,30 @@ export const cleanComponentMocks = (): void => {
 };
 
 export const emptyResponseProvider: ResponseProvider = () => ({});
+export const submitButtonSelector: ElementSelector = {
+  querySelector: 'button[type="submit"]',
+};
+export const getElementAttribute = (
```
VilleMiekkoja: Not sure if this function brings extra value? Could we just use element.getAttribute directly?

src/common/test/componentMocking.tsx
```diff
@@ -231,27 +256,29 @@ export const renderComponentWithMocksAndContexts = async (
     }
   };
 
+  const waitForElementAndValue: TestTools['waitForElementAndValue'] = async props => {
+    const { selector, value } = props;
+    return waitFor(async () => {
+      const elementValue = await getTextOrInputValue(selector);
+      if (value && (!elementValue || !elementValue.includes(value))) {
```
VilleMiekkoja: What if `value` is false in this condition? It seems to wait for elementValue?

src/profile/components/basicData/__tests__/BasicData.test.tsx
```diff
@@ -181,4 +184,75 @@ describe('<BasicData /> ', () => {
...
+        // set invalid values
+        await setValues(setInputValue, invalidData);
+        // submit also validates the form
+        await clickElement(submitButtonSelector);
+        await waitForElementAttributeValue(
+          elementGetter,
+          'aria-invalid',
+          'true'
+        );
+        // getElement throws is element is not found
```
VilleMiekkoja: typo: is -> if

src/profile/components/basicData/__tests__/BasicData.test.tsx
```diff
@@ -181,4 +184,75 @@ describe('<BasicData /> ', () => {
...
+        // so element is not checked in expect
+        errorElementGetter();
```
VilleMiekkoja: This probably works already as it is, but could we use this here in addition:

`expect(() => ...error...).not.toThrow(error)`?

src/profile/hooks/useProfileDataEditor.ts
```diff
@@ -96,14 +102,31 @@ export function useProfileDataEditor({
...
+    triggerUpdate();
+    return Promise.resolve();
+  };
+
   return {
     editDataList: getEditData(),
     save,
     reset,
+    add,
+    hasNew,
```
VilleMiekkoja, line 127: Could we inline the hasNewItem here?

src/profile/helpers/formik.ts
```diff
@@ -51,3 +51,24 @@ export function getFieldError<FormValues>(
     t(message, options);
   return getError<FormValues>(formikProps, fieldName, renderError, isOldData);
 }
+
+export function createFormFieldHelpers<FormikValues>(
+  t: TFunction,
```
VilleMiekkoja, line 56: In general I think variables with only 1 letter should be avoided, since they are not very descriptive and they are more harder to find with text editors (for example ctrl + f). IDEs can help though. I guess this is the translate function (t) of the i18next library? Probably not much then what could be done here, as it might be confusing to use another name for it.

src/profile/helpers/editData.ts
```diff
@@ -115,11 +118,15 @@ function isSaving(allItems: EditData[]): boolean {
...
 }
 
-function getNewItem(allItems: EditData[]): EditData | undefined {
-  return allItems.find(item => item.id === '');
+export function isNewItem(data: EditData): boolean {
+  return data.id === '';
 }
 
-function hasNewItem(allItems: EditData[]): boolean {
+export function getNewItem(allItems: EditData[]): EditData | undefined {
```
VilleMiekkoja, line 125: Is there always only 1 new item in the array? Could it be more safe to rename this: getFirstNewItem?

src/profile/helpers/editData.ts
```diff
@@ -103,9 +103,12 @@ export type EditFunctions = {
   updateData: (newProfileRoot: ProfileRoot) => boolean;
   updateAfterSavingError: (id: string) => boolean;
   resetItem: (targetItem: EditData) => boolean;
+  addItem: () => EditData;
+  hasNewItem: () => boolean;
+  removeItem: (targetItem: EditData) => Partial<FormValues> | null;
```
VilleMiekkoja, line 108: Why do we wish to sometimes return partial FormValues and sometimes return null?

src/profile/components/multiItemRow/multiItemRow.module.css
```diff
@@ -0,0 +1,43 @@
...
+.rowContentWrapper form {
+  width: 100%;
+}
+
+.editableRow {
+  display: flex;
+  flex-direction: column;
+}
+
+/* fixes bug in HDS TextInput when it is wrapped to an element with display:flex: */
```
VilleMiekkoja, line 16: There might be a fix coming to this in the next release of HDS 😅 . We removed TextInput input wrapper height 100%, which might be related to this.

src/profile/components/multiItemRow/multiItemRow.module.css
```diff
@@ -0,0 +1,43 @@
...
+  background: var(--color-black-ultra-light);
+  margin: 0 0 var(--spacing-m) 0;
+  padding: var(--spacing-m);
+}
+
+.rowContentWrapper form {
+  width: 100%;
+}
+
+.editableRow {
```
VilleMiekkoja, line 11: Airbnb styleguide prefers dashes over camelCasing in class names: https://github.com/airbnb/css#css

dev.hel.fi recommends following the airbnb's style guide with some exceptions: https://dev.hel.fi/coding-standards#reactjsx-style-guide

Doesn't refer to css there though, so not sure if it should be followed in styles also?

src/profile/components/multiItemRow/multiItemRow.module.css
```diff
@@ -0,0 +1,43 @@
...
+  flex-direction: column;
+}
+
+/* fixes bug in HDS TextInput when it is wrapped to an element with display:flex: */
+.editableRow > div > div {
+  height: auto;
+}
+
+.value {
+  font-size: 18px;
```
VilleMiekkoja, line 22: Could we apply the hds design tokens here? https://hds.hel.fi/design-tokens/typography#tokens. I think it would be the `$fontsize-body-l: 1.125rem; // 18px`

src/profile/hooks/useCommonEditHandling.ts
```diff
@@ -0,0 +1,53 @@
...
+  isNew: boolean;
+  actionHandler: ActionHandler;
+};
+
+export const useCommonEditHandling = (
+  props: RowItemProps
+): UseActionHandlingReturnType => {
+  const { data, onAction } = props;
+  const isNew = isNewItem(data);
+  const [isEditing, setEditing] = useState(isNew);
```
VilleMiekkoja, line 21: isNew means it is isEditing? Asking because it is passed as the default value here

src/profile/hooks/useCommonEditHandling.ts
```diff
@@ -0,0 +1,53 @@
...
+  const [currentAction, setCurrentAction] = useState<Action>(undefined);
+  const actionHandler: ActionHandler = async (action, newValue) => {
+    if (action === 'set-primary' || action === 'remove' || action === 'save') {
+      setCurrentAction(action);
+    }
+    const [err] = await to(onAction(action, data, newValue as EditDataValue));
+    if (isNew) {
+      if (err) {
+        setCurrentAction(undefined);
```
VilleMiekkoja, line 30: Should we also reject on error? Since there's no return inside this if, the actionHandler will resolve on line 32 even in the case of error?

src/profile/components/multiItemRow/MultiItemRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+    isEditing,
+    currentAction,
+    actionHandler,
+    isNew,
+  } = useCommonEditHandling(props);
+  const { value, primary } = data;
+  const schema = dataType === 'phones' ? phoneSchema : emailSchema;
+  const inputValue: string =
+    (dataType === 'phones'
+      ? (value as PhoneValue).phone
```
VilleMiekkoja, line 37: It's little hard to follow quickly where value comes from. Would maybe prefer applying full path like: `props.data.value` that shows it right away that it comes from props. Or then try to destructure it only once from props.

src/profile/components/multiItemRow/MultiItemRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+          validationSchema={schema}
+        >
+          {(formikProps: FormikProps<EmailAndPhoneFormikValue>) => (
+            <Form>
+              <div className={styles.editableRow}>
+                <Field
+                  name="value"
+                  id={inputId}
+                  maxLength={formFields.value.max as number}
+                  as={TextInput}
```
VilleMiekkoja, line 79: Should we apply the HDS PhoneInput component here when type is phone? https://hds.hel.fi/components/phone-input

One benefit is that it adds the type=tel to the input, which then helps mobile users by opening the number keyboard

src/profile/components/multiItemRow/MultiItemRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+import EditButtons from '../editButtons/EditButtons';
+import commonFormStyles from '../../../common/cssHelpers/form.module.css';
+import SaveIndicator from '../saveIndicator/SaveIndicator';
+import { useCommonEditHandling } from '../../hooks/useCommonEditHandling';
+import { RowItemProps } from '../multiItemEditor/MultiItemEditor';
+import { getFormFields } from '../../helpers/formProperties';
+
+type EmailAndPhoneFormikValue = { value: string };
+
+function MultiItemRow(props: RowItemProps): React.ReactElement {
```
VilleMiekkoja, line 21: I fear there can be a risk that combining phone and email to same row component might not scale in the long run in terms of readability of the code. It makes code more abstractly named, and there's going to be places where you have to check the data-type, which increases the complexity.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+              {primary
+                ? t('profileInformation.primaryAddress')
+                : t('profileInformation.address')}
+            </h4>
+            <div className={commonFormStyles.multiItemWrapper}>
+              <Field
+                name="address"
+                id={`${testId}-address`}
+                maxLength={formFields.address.max as number}
+                as={TextInput}
```
VilleMiekkoja, line 78: Could we apply HDS NumberInput component here? https://hds.hel.fi/components/number-input

It aids screen readers and gives better mobile keyboard

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+                helperText={getFieldErrorMessage(formikProps, 'address')}
+                labelText={t(formFields.address.translationKey)}
+                autoFocus
+                aria-labelledby={`${dataType}-address-helper`}
+              />
+              <Field
+                name="postalCode"
+                id={`${testId}-postalCode`}
+                maxLength={formFields.postalCode.max as number}
+                as={TextInput}
```
VilleMiekkoja, line 100: Could we apply HDS NumberInput component here? https://hds.hel.fi/components/number-input

It aids screen readers and gives better mobile keyboard

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -98,6 +102,32 @@ describe('editData.ts ', () => {
...
+      newItem,
+      oldIdList,
+    };
+  };
+  const pickMultiItemFormValues = (
+    formValues: FormValues | Partial<FormValues>,
+    dataType: EditDataType
+  ): MultiItemProfileNode[] =>
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    //@ts-ignore
```
VilleMiekkoja, line 128: I wonder if there's a way to type the row 129 without typescript warnings?

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -172,16 +202,16 @@ describe('editData.ts ', () => {
         updateItemAndCreateSaveData,
```
VilleMiekkoja, line 202: Off-topic: comment for line 193: my IDE suggest `!!node.primary` could be simplified to `node.primary`

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+type FormikValues = AddressValue;
+
+function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
+  const { data, testId, dataType } = props;
+  const value = data.value as AddressValue;
+  const { address, city, postalCode, countryCode } = value;
+  const { t, i18n } = useTranslation();
+  const lang = i18n.languages[0];
+  const applicationLanguage = getLanguageCode(i18n.languages[0]);
+  const countryList = countries.getNames(applicationLanguage);
```
VilleMiekkoja, line 34: Would it be good idea to sort the countries based on names? Most of them seemed to be sorted already, but I found in the end of the list there's countries that are not sorted with the rest of the countries?
image

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+                aria-invalid={hasFieldError(formikProps, 'address')}
+                helperText={getFieldErrorMessage(formikProps, 'address')}
+                labelText={t(formFields.address.translationKey)}
+                autoFocus
+                aria-labelledby={`${dataType}-address-helper`}
+              />
+              <Field
+                name="postalCode"
+                id={`${testId}-postalCode`}
+                maxLength={formFields.postalCode.max as number}
```
VilleMiekkoja, line 99: Tested this and the maxLength seemed to be 5? I think there might be longer postal codes in other countries, so should we support longer ones as well, because we want to support other countries too?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'city')}
+                aria-invalid={hasFieldError(formikProps, 'city')}
+                helperText={getFieldErrorMessage(formikProps, 'city')}
+                labelText={t(formFields.city.translationKey)}
+                aria-labelledby={`${dataType}-city-helper`}
+              />
+              <div className={commonFormStyles.formField}>
+                <FormikDropdown
+                  name="countryCode"
```
VilleMiekkoja, line 120: If the country is USA, and we wish to support USA addresses, should we add state field also, when USA is selected? Im not sure if state is compulsory for USA address? 😅

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -254,6 +288,234 @@ describe('editData.ts ', () => {
...
+        expect(newItem.saving).toBeUndefined();
+        if (dataType === 'emails') {
+          expect((newItem.value as EmailValue).email).toBeDefined();
+        } else if (dataType === 'phones') {
+          expect((newItem.value as PhoneValue).phone).toBeDefined();
+        } else if (dataType === 'addresses') {
+          expect((newItem.value as AddressValue).address).toBeDefined();
+        }
+      });
+      it(`updateItemAndCreateSaveData updates the new ${dataType} item and it is in returned formValues`, () => {
```
VilleMiekkoja: I wonder if the language of this sentence could be improved? Im not sure I got the ending part of it?

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -254,6 +288,234 @@ describe('editData.ts ', () => {
...
+        expect(saveData).toBeDefined();
+        expect(updatedItem.saving).toEqual('remove');
+        // original item is not mutated
+        expect(removeItem.saving).toBeUndefined();
+        const itemsInFormValues = pickMultiItemFormValues(saveData, dataType);
+        // item at target index has chaged
+        expect(itemsInFormValues[testIndex].id).toEqual(
+          editDataWithRemovedItem[testIndex + 1].id
+        );
+        // edit datas have not changed yet
```
VilleMiekkoja: edit datas -> edit data

src/profile/components/multiItemEditor/MultiItemEditor.tsx
```diff
@@ -0,0 +1,173 @@
...
+        className={commonFormStyles.responsiveButton}
+        id={`${dataType}-add-button`}
+      >
+        {texts.addNew}
+      </Button>
+    </ProfileSection>
+  );
+}
+
+export default MultiItemEditor;
```
VilleMiekkoja, line 173: Should we use the inline export instead of this?

src/profile/components/multiItemEditor/MultiItemEditor.tsx
```diff
@@ -0,0 +1,173 @@
...
+    setErrorMessage,
+    setSuccessMessage,
+    clearMessage,
+  } = useNotificationContent();
+
+  const hasAddressList = dataType === 'addresses';
+  const isAddButtonDisabled = hasNew();
+  const RowComponent = hasAddressList ? MultiItemAddressRow : MultiItemRow;
+  const texts = (function() {
+    if (dataType === 'emails') {
```
VilleMiekkoja, line 58: Hmm.. I wonder if there's a risk that this file becomes too complex, since we are trying to support all different dataTypes? If the requirements of different data types start changing a lot, it might be a good idea to apply some refactoring.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+          disabled={!!currentAction}
+          testId={testId}
+        />
+        <SaveIndicator action={currentAction} testId={testId} />
+      </div>
+    </div>
+  );
+}
+
+export default MultiItemAddressRow;
```
VilleMiekkoja, line 199: Should we change this export to inline?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
+import { TextInput } from 'hds-react';
```
VilleMiekkoja, line 1: Looking good! I wonder if the name of the file could be simplified to: AddressRow?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+                maxLength={formFields.city.max as number}
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'city')}
+                aria-invalid={hasFieldError(formikProps, 'city')}
+                helperText={getFieldErrorMessage(formikProps, 'city')}
+                labelText={t(formFields.city.translationKey)}
+                aria-labelledby={`${dataType}-city-helper`}
+              />
+              <div className={commonFormStyles.formField}>
+                <FormikDropdown
```
VilleMiekkoja, line 119: EDIT: was about to suggest applying HDS Dropdown, but then noticed that the FormikDropdown applies it.

Combobox might fit better, since there's more than 8 options, and it might be quicker for the user to input first letter of the country and get the result?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
VilleMiekkoja, line 120: Should this be country instead of countryCode?

src/profile/components/multiItemEditor/__tests__/MultiItemEditor.test.tsx
```diff
@@ -0,0 +1,362 @@
...
+            id: `${dataType}-${index}-${
+              dataType === 'addresses' ? field : 'value'
+            }`,
+          },
+          newValue: value,
+        });
+      }
+    };
+
+    describe(`Renders ${dataType} and fields of that
```
VilleMiekkoja: I wonder if the language could be improved here hmm.. _Renders address type and its fields:_?

src/profile/components/multiItemEditor/__tests__/MultiItemEditor.test.tsx
```diff
@@ -0,0 +1,362 @@
...
+            },
+          });
+          await waitFor(() => {
+            expect(isDisabled(addButton)).toBeFalsy();
+          });
+          await verifyValues(getTextOrInputValue, formValues, newItemIndex);
+        });
+      });
+
+      it(`Component shows error notification and stays in edit mode when error occures.
```
VilleMiekkoja: Typo: occures -> occurs

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <div className={commonFormStyles.multiItemWrapper}>
+              <Field
+                name="address"
+                id={`${testId}-address`}
+                maxLength={formFields.address.max as number}
+                as={TextInput}
+                invalid={hasFieldError(formikProps, 'address')}
+                aria-invalid={hasFieldError(formikProps, 'address')}
+                helperText={getFieldErrorMessage(formikProps, 'address')}
+                labelText={t(formFields.address.translationKey)}
```
VilleMiekkoja, line 92: The font-weight and size of these input labels seem to change depending on mode (edit or view). It might be what is wanted, but at first glimpse looked little weird.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -0,0 +1,189 @@
...
+        validationSchema={addressSchema}
+      >
+        {(formikProps: FormikProps<FormikValues>) => (
+          <Form className={commonFormStyles.multiItemForm}>
+            <h4 className={commonFormStyles.sectionTitle}>
+              {primary
+                ? t('profileInformation.primaryAddress')
+                : t('profileInformation.address')}
+            </h4>
+            <div className={commonFormStyles.multiItemWrapper}>
```
VilleMiekkoja, line 83: The email and phone rows have a gray background color, I wonder if we would want the address row to have it too?

image

VilleMiekkoja:
Tested this locally, and seems to be working great! Good job with this 🙌 .

src/profile/components/multiItemRow/multiItemRow.module.css
VilleMiekkoja: The fix should now be available: https://www.npmjs.com/package/hds-react

src/common/test/componentMocking.tsx
NikoHelle: Yep. 👍   

Could split types, helpers and rendering?

src/common/test/componentMocking.tsx
NikoHelle: Yes, there used to be check for element existence in the func, but not anymore. I'll remove the func.

src/common/test/componentMocking.tsx
NikoHelle: Value cannot be false, it must be a string. "Value" is inside the "if" to cover the case when it is an empty string ''.

This is quite unclear and I needed some time to find out why it is there. I changed that part to be a lot clearer:

if (value === '' && elementValue === '') {
        return;
      }

ElementValue can be "undefined" until element is found in the dom. That's why !elementValue wont work.

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: This is probably clearer

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Do you mean adding 

`hasNew: ()=>hasNewItem?`
 
It is not allowed by sonarCube, I think. It wont allow mixing shorthands with other code in "return {...}" blocks.

But I changed the code where hasNewItem is created to 

`hasNewItem: hasNew,`

so "const hasNew = () => hasNewItem();"  is removed.

src/profile/helpers/formik.ts
NikoHelle: Yes, "t" it is **the** i18n.t, so kept the same name with the function. One letter funcs/vars are bad, but in this case it clarifies that it is **the** "t" -function.

src/profile/helpers/editData.ts
NikoHelle: There can be only one (or none), so no need to clarify it.

src/profile/helpers/editData.ts
NikoHelle: When formValues are returned the caller knows it needs to save something. Null is returned when a new, unsaved, item is removed and there is no need for server call.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: NumberInput, PhoneInput and other HDS fixes mentioned here were released after this was originally created. HDS update would probably cause a lot of changes, so did not want to update it in this PR and make it larger. I'd rather make a new ticket for HDS updates.

src/profile/components/multiItemRow/MultiItemRow.tsx
NikoHelle: NumberInput, PhoneInput and other HDS fixes mentioned here were released after this was originally created. HDS update would probably cause a lot of changes, so did not want to update it in this PR and make it larger. I'd rather make a new ticket for HDS updates.

src/profile/components/multiItemRow/multiItemRow.module.css
NikoHelle: I have kept the same naming that all other css files follow in the Profile UI. Original files are probably older than the spec in dev.hel.fi. Maybe all files should be changed in one update or just this one now?

Maybe combine this to a HDS update.

src/profile/components/multiItemRow/multiItemRow.module.css
NikoHelle: NumberInput, PhoneInput and other HDS fixes mentioned here were released after this was originally created. HDS update would probably cause a lot of changes, so did not want to update it in this PR and make it larger. I'd rather make a new ticket for HDS updates.

src/profile/hooks/useCommonEditHandling.ts
NikoHelle: No, if the item is a new one, then edit mode should always be "active".

src/profile/hooks/useCommonEditHandling.ts
NikoHelle: Awaiting function does not care what happens, because all error handling is here. Adding rejection would also require rejection handling to the awaiting function, because all promise rejections must be handled.

src/profile/components/multiItemRow/MultiItemRow.tsx
NikoHelle: Doesn't this apply to all destructured variables? So never use it? I'm a bit lost how to solve this and make it clearer:
```
const { data, testId, dataType } = props;
...
const { value, primary } = data;
```

src/profile/components/multiItemRow/MultiItemRow.tsx
NikoHelle: IMHO, data-type checks are easier to follow than making two files (EmailRow and PhoneRow) and then more files for common code of those two. The need for scaling this file is an abstract scenario. If there are new features, it is easier to add them here?

Currently there are 3 places where dataType is checked and 2 where it is passed as argument.
There are 122 lines of code and only those 3 lines would not be common if split in to multiple files.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Yes, if changed to this:

```
const pickMultiItemFormValues = (
    formValues: FormValues | Partial<FormValues>,
    dataType: EditDataType
  ): MultiItemProfileNode[] => {
    if (
      dataType === 'addresses' ||
      dataType === 'emails' ||
      dataType === 'phones'
    ) {
      return formValues[dataType] as MultiItemProfileNode[];
    }
    throw new Error('Invalid datatype for picking MultiItemProfileNodes');
  };
```

I liked the previous version because this is a test file.

src/profile/components/multiItemRow/multiItemRow.module.css
VilleMiekkoja: Yeah it might be a big task to change them all, so not sure is it worth it. Though if we're still in the beginning part of creating the open-city-profile-ui, it might be worth it.

src/profile/components/multiItemRow/MultiItemRow.tsx
VilleMiekkoja: Yeah it probably does 😅 . I just think that if we get a variable by destructuring multiple times at different rows, it makes it little harder to read where it originally came from. Is it possible to get it by destructuring it only once? Or then just apply the props.data.value directly

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: What language is your browser or profile? I've different ordering.

NikoHelle: If UI language is English, the sorting is messed up. In Finnish version, only "Kosovo" is in wrong place. Ill' sort them manually.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: There is no documentation why 5 was selected in old Profile UI. Perhaps, because BE didn't always support more chars. I asked around and now BE supports 32 chars. UX designer approved this too. So will change to 32.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Good point. Currently BE has no support for US states or whatever China or Zimbabwe needs 😃  This needs additional designing.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: I chose "MultiItemAddressRow" so it would group with "MultiItemRow" and "MultiItemEditor"

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Following other exports in .tsx-files here.

src/profile/components/multiItemEditor/MultiItemEditor.tsx
NikoHelle: Following other exports in .tsx-files here

src/profile/components/multiItemEditor/MultiItemEditor.tsx
NikoHelle: Actually supporting only 3 data types. :) Those who have multiple items and have same editing tools and same notifications.

All this does it handle identical actions (save, add ,remove, set-primary), show notifications and render each item.

NikoHelle: Multiple items need a "parent / wrapper" that knows what children are doing. Later on, it must prevent children from being edited while re-ordering is in progress. Now it prevents multiple new items. 

"If the requirements of different data types start changing a lot, "

Then it should be addressed, perhaps not now?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Currently you can type a letter and it jumps there. I don't know what HDS combobox is and when it was introduced. The designs do not use it.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: BE, profileData and all other places us countryCode.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: That is not in the layouts, so probably not wanted.

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Looks like HDS and layouts have different styles. HDS should be followed. Another fix to be done in the HDS update....

src/profile/components/multiItemRow/multiItemRow.module.css
NikoHelle: Yes, HDS should be followed. Another fix to be done in the HDS update....

src/profile/components/multiItemEditor/__tests__/MultiItemEditor.test.tsx
NikoHelle: Yes, I'll translate it to English :)

src/profile/components/multiItemEditor/MultiItemEditor.tsx
VilleMiekkoja: Ok, sorry I thought that inline export was the convention.

src/common/test/componentMocking.tsx
NikoHelle: Done.

src/common/test/componentMocking.tsx
NikoHelle: Done.

src/common/test/componentMocking.tsx
NikoHelle: Done.

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: Done.

src/profile/components/basicData/__tests__/BasicData.test.tsx
NikoHelle: Done.

src/profile/hooks/useProfileDataEditor.ts
NikoHelle: Done.

src/profile/components/multiItemRow/multiItemRow.module.css
NikoHelle: There wont be any new css files :)

src/profile/components/multiItemRow/MultiItemRow.tsx
NikoHelle: Changed to const { data:{ value , primary }, testId, dataType } = props;

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Done.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: changed. Required a new commit, because this wasn't added in this PR

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Added sorting

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
NikoHelle: Done

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Fixed.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Fixed.

src/profile/components/multiItemEditor/__tests__/MultiItemEditor.test.tsx
NikoHelle: Fixed.

src/profile/components/multiItemEditor/__tests__/MultiItemEditor.test.tsx
NikoHelle: Fixed.

src/profile/components/multiItemRow/multiItemRow.module.css
NikoHelle: HDS update ticket created https://helsinkisolutionoffice.atlassian.net/browse/HP-727

VilleMiekkoja (approved):

PR #164, opened 2021-05-18
HP-318 step 8: set primary

NikoHelle:
Addresses, emails and phones can now be set as primary. While changing the primary item, all other modifications are prohibited by disabling buttons in that group.

Primary item is always the first item in UI. Backend orders only emails like this, so at the moment UI re-orders them, if primary is not at index 0.

### DIFF
--- a/src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
+++ b/src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
@@ -25,7 +25,7 @@ import { RowItemProps } from '../multiItemEditor/MultiItemEditor';
 type FormikValues = AddressValue;
 
 function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
-  const { data, testId, dataType } = props;
+  const { data, testId, dataType, disableEditButtons } = props;
   const value = data.value as AddressValue;
   const { address, city, postalCode, countryCode } = value;
   const { t, i18n } = useTranslation();
@@ -57,7 +57,8 @@ function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
     FormikValues
   >(t, isNew);
 
-  const { primary } = data;
+  const { primary, saving } = data;
+  const disableButtons = !!currentAction || !!saving;
 
   if (isEditing) {
     return (
@@ -133,7 +134,7 @@ function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
             </div>
             <FormButtons
               handler={actionHandler}
-              disabled={!!currentAction}
+              disabled={disableButtons}
               testId={testId}
               alignLeft
             />
@@ -183,11 +184,11 @@ function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
           actions={{
             removable: !primary,
             primary,
-            setPrimary: false,
+            setPrimary: true,
           }}
           buttonClassNames={commonFormStyles.actionsWrapperButton}
           editButtonId={`${testId}-edit-button`}
-          disabled={!!currentAction}
+          disabled={disableButtons || disableEditButtons}
           testId={testId}
         />
         <SaveIndicator action={currentAction} testId={testId} />
--- a/src/profile/helpers/__tests__/editData.test.ts
+++ b/src/profile/helpers/__tests__/editData.test.ts
@@ -22,6 +22,8 @@ import {
   pickValue,
   saveTypeSetPrimary,
   updateItems,
+  setPrimary,
+  movePrimaryAsFirst,
 } from '../editData';
 import {
   cloneProfileAndProvideManipulationFunctions,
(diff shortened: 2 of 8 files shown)


src/profile/helpers/editData.ts
```diff
@@ -334,6 +339,51 @@ export function updateItems(
...
+// remove this when backend orders them
+export function movePrimaryAsFirst(allItems: EditData[]): EditData[] {
+  const primaryItemIndex = allItems.findIndex(item => item.primary);
+  if (primaryItemIndex < 1) {
+    return allItems;
+  }
+  const primaryItem = allItems[primaryItemIndex];
+  const firstItem = allItems[0];
+  allItems[0] = primaryItem;
+  allItems[primaryItemIndex] = firstItem;
```
akikoskinen: Instead of swapping, should the primary just be moved to first, and let the rest be in the order they were in?

src/profile/components/multiItemAddressRow/MultiItemAddressRow.tsx
```diff
@@ -57,7 +57,8 @@ function MultiItemAddressRow(props: RowItemProps): React.ReactElement {
     FormikValues
   >(t, isNew);
 
-  const { primary } = data;
+  const { primary, saving } = data;
+  const disableButtons = !!currentAction || !!saving;
```
NikoHelle, line 62: The "currentAction" is not the only indicator for item being saved. "saving" property can also change when item is primary and another item is being set as primary.

src/profile/helpers/editData.ts
```diff
@@ -334,8 +341,54 @@ export function updateItems(
   return dataHasUpdated ? newList : null;
 }
 
+// this is a temporary function to assure primaryItem is always at index 0
+// remove this when backend orders them
```
akikoskinen: Better remove this comment. I started thinking about the ordering of these lists on the server side. I may be of the opinion that they shouldn't actually be ordered (at least by default) according to the `primary` attribute, but according to something stable, like an id. The reason is that the server provides the list as a [Relay Connection](https://relay.dev/graphql/connections.htm) and in such a pagination model the ordering should be stable.

src/profile/helpers/__tests__/editData.test.ts
```diff
@@ -540,14 +605,53 @@ describe('editData.ts ', () => {
...
+      const [initialPrimary, nextPrimary] = editDataList;
+      // items are not being saved yet, so function returns false indication nothing to update
+      const currentPrimaryIsNotRolledBack = updateAfterSavingError(
+        initialPrimary.id
+      );
+      const nextPrimaryIsNotRolledBack = updateAfterSavingError(nextPrimary.id);
+      expect(currentPrimaryIsNotRolledBack).toBeFalsy();
+      expect(nextPrimaryIsNotRolledBack).toBeFalsy();
+
+      // imported setPrimary -function not used to keep data in scope with updateAfterSavingError and getEditData
```
akikoskinen: I can't understand this comment. What is it trying to communicate?

akikoskinen (commented):
Few comments, nothing major.

src/profile/helpers/editData.ts
NikoHelle: Order by id is fine. UI will anyway put the primary item as first. Items should be ordered by some attribute/value -pair that is returned in GraphQL, so UI can order them too after primary has changed.  This way UI can order them immediately without waiting for backend.

id is a string. Is it alphabetically sortable and newest is last in A-Z ordering? It would be confusing for the user if newest would jump to other position than last.

src/profile/helpers/editData.ts
NikoHelle: Swapping matches quite well what backend returns after save. I know there is no actual sorting done on the backend. Except in emails primary is the first one. Some "groups" (addresses, phones, emails) seem to swap, some just push new primary as first.The discussion below (https://github.com/City-of-Helsinki/open-city-profile-ui/pull/164#discussion_r638519217) will hopefully result in some logic.

I've discussed this with UX-designers and it is fine to swap items even though it does not match save result every time. Re-ordering after save could only occur when user has 3 or more items.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Most of the tests in this file use the imported setPrimary-function that is not bound to any scope or therefore any items. This test uses the setPrimary returned from createEditorForDataType() -call. Its scope is same as other functions in this test.

akikoskinen: Would this comment be more helpful at the place where the `createEditorForDataType()` call is?

src/profile/helpers/editData.ts
akikoskinen: > Order by id is fine. UI will anyway put the primary item as first. Items should be ordered by some attribute/value -pair that is returned in GraphQL, so UI can order them too after primary has changed. This way UI can order them immediately without waiting for backend.
> 
> id is a string. Is it alphabetically sortable and newest is last in A-Z ordering? It would be confusing for the user if newest would jump to other position than last.

In the database the `id` is numeric and increases for new items so that's a natural stable key for sorting. The `id` returned via GraphQL is based on the database `id`: it's prefixed with some static string and then base64 encoded. I guess mostly sorting those alphabetically leads to the same result, but there could be cases where that doesn't quite work. But it's not guaranteed that the GraphQL `id` calculation algorithm will remain as described here. It's an implementation detail and shouldn't be relied on.

But does UI need to do full sorting? It has the list it gets from server. It shows the primary item as first but otherwise the list order is the same. When primary item is changed, the used-to-be-primary item returns to its original place in the UI and the new primary gets to be shown first. That should match the order received from server, right?

src/profile/helpers/editData.ts
akikoskinen: Yes, besides the primary being first with emails, there's no sorting in backend. They come in whatever order the database happens to return them. This should be fixed in the backend, so that the order is stable and deterministic.

There's no point trying to mimic the current backend's order in UI, as the order in backend is currently basically random. So whatever might seem to work in some case, fails in others, or the next day.

src/profile/helpers/editData.ts
NikoHelle: Sounds good, when sorted with id! UI has to either save the original order of ids or keep the ordering and re-order primary only when data is rendered. The latter sounds better at the moment.

I'll add this here right now., but backend has to be changed too. Before backend sorts by "id", there is some glitches in UI when items change place. But that won't show unless user has 3 or more items.

src/profile/helpers/__tests__/editData.test.ts
NikoHelle: Could be. I'll move it up.

src/profile/helpers/editData.ts
NikoHelle: The other conversation about ordering made this irrelevant as code will change.

src/profile/helpers/editData.ts
akikoskinen: Yes, putting primary first at the "very last moment" in the UI sounds sensible to me too. So it's just for visualization. Everywhere else in the UI project code the order stays unchanged.

akikoskinen (approved):

PR #166, opened 2021-05-28
HP-318 step 9: change language and improve accessibility

NikoHelle:
Language can be changed and saved to Profile API.

Added accessibility improvements:
- when a form is closed, focus is returned to the button that opened a form
- aria-labels of edit buttons have better description of the target
- focus is trapped inside a form until it is saved

Also deleted unused files


src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+describe('<focusTrapper /> ', () => {
+  it('keeps focus inside', async () => {
+    let focusEventCount = 0;
+    const focusListener = (e: React.FocusEvent<HTMLInputElement>) => {
+      const id = e.target.getAttribute('id');
+      if (id === 'target') {
+        focusEventCount = focusEventCount + 1;
+      }
+    };
```
akikoskinen: About names: instead of a general `focusListener` function name, this could have a more descriptive name, maybe something like `incrementTargetElementFocusCounter`. Similarly, since `focusEventCount` seems to count only `target` element's focuses, it could be called e.g. `targetElementFocusCount`.

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const result = render(
+      <div>
+        <FocusKeeper targetId={'target'} className="focus-keeper">
+          <input type="text" id="target" onFocus={focusListener} />
+          <input type="text" id="second" onFocus={focusListener} />
```
akikoskinen: Since `focusListener` counts only events for the `target` element, it doesn't have to be added to the `second` element?

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const getTrapperElement = (pos: 'above' | 'below') => {
+      const focusKeeperSelector = 'div.focus-keeper';
+      const trapperSelector =
+        pos === 'above' ? '> div:first-child' : '> div:last-child';
+      return result.container.querySelector(
+        `${focusKeeperSelector} ${trapperSelector}`
+      ) as HTMLElement;
+    };
+    const focusTrapperAbove = getTrapperElement('above');
+    const focusTrapperBelow = getTrapperElement('below');
```
akikoskinen: IMO the "above"/"below" terminology is somewhat obscure. "Above" (or "below") what? The CSS selectors seem to provide better terms: "first child", "last child". That terminology could be used instead of "above" and "below".

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    // move focus outside of FocusKeeper
+    focusTrapperAbove.focus();
+    result.container.focus();
+    focusTrapperBelow.focus();
+    expect(focusEventCount).toEqual(2);
```
akikoskinen: Should the focus count be checked after each step, to ensure that the automatic focus changing works in the expected way?

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      const trapperSelector =
+        pos === 'above' ? '> div:first-child' : '> div:last-child';
+      return result.container.querySelector(
+        `${focusKeeperSelector} ${trapperSelector}`
+      ) as HTMLElement;
+    };
+    const focusTrapperAbove = getTrapperElement('above');
+    const focusTrapperBelow = getTrapperElement('below');
+    // move focus outside of FocusKeeper
+    focusTrapperAbove.focus();
```
akikoskinen: This comment seems a bit misplaced, as the following line seems to do exactly the opposite, moving the focus **inside** the `FocusKeeper`. Perhaps there doesn't have to be a comment here at all.

src/profile/components/additionalInformation/AdditionalInformation.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+  const editData = editDataList[0];
+  const { value } = editData;
+  const { language } = value as AdditionalInformationValue;
+  const languageOptions: OptionType[] = profileConstants.LANGUAGES.map(
+    languageOption => ({
+      value: languageOption,
+      label: t(`LANGUAGE_OPTIONS.${languageOption}`),
+    })
+  );
```
akikoskinen, line 47: Is this a static value? Would it make sense to construct it outside the component function and then just use it here?

src/profile/components/additionalInformation/AdditionalInformation.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <ProfileSection>
+      <h3 className={commonFormStyles.sectionTitle}>
+        {t('profileForm.language')}
+      </h3>
+      <div className={commonFormStyles.multiItemWrapper}>
+        <Formik
+          initialValues={{ language }}
+          onSubmit={() => {
+            /* onChange does this. This comment is here because onSubmit is mandatory and function cannot be empty */
```
akikoskinen, line 73: Kind of weird, that every other change in the data requires a "start modifying/modify/save" cycle, but this one field is saved immediately :thinking:

src/profile/components/additionalInformation/__tests__/AdditionalInformation.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  const initTests = async (): Promise<TestTools> => {
+    responses.push({ profileData: initialProfile });
+    const testTools = await renderTestSuite();
+    await testTools.fetch();
+    return Promise.resolve(testTools);
+  };
+
+  it('should render language selection, auto save and show success/error message', async () => {
```
akikoskinen: This one test has quite a many responsibilities. It seems that it could be split at least into two, the success and error cases.

src/profile/components/additionalInformation/AdditionalInformation.tsx
akikoskinen: Also, IMO this leads to a bit sub-par UX in error situations. With the current solution, when one selects another language, but the saving fails, the newly selected language remains selected, and an error is shown which asks to "try again". How does the user try again when there's no save button? The user has to temporarily select a third language, then reselect the one they originally wanted to select and hope that this time the saving succeeds.

I wonder if this is the best flow for the user...

src/profile/helpers/createActionAriaLabels.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (dataType === 'addresses') {
+    return {
+      edit: `${t('profileInformation.ariaEditAddress')} ${value}`,
+      primary: `${t('profileInformation.primaryAddressAriaDescription')}`,
+      setPrimary: `${t('profileForm.setPrimary')} ${t(
+        'profileInformation.address'
+      )} ${value}`,
+      remove: `${t('profileForm.remove')} ${t(
+        'profileInformation.address'
+      )} ${value}`,
```
akikoskinen: Data should be inserted into the translations with place holders, not by concatenating strings in code. Nor should translated strings be concatenated together to form new user visible strings.

akikoskinen (changes requested):

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
NikoHelle: Second listener was obsolete, so removed it. Kept same function name, because it is a focusListener (now for single element), but changed focusEventCount to focusEventsOnTarget

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
NikoHelle: Removed

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
NikoHelle: Changed above -> before, below -> after and added better explanation which is which.

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
NikoHelle: Thx! Added more checks

src/common/focusKeeper/__tests__/FocusKeeper.test.tsx
NikoHelle: Yeah, moved it down to correct line.

src/profile/components/additionalInformation/AdditionalInformation.tsx
NikoHelle: It changes when UI language changes, so it is not static.

src/profile/components/additionalInformation/AdditionalInformation.tsx
NikoHelle: Good point and clearly a bug. The selected value is now reverted when an error occurs, so change can be made again without doing any dummy changes.

IMO, this is much more convenient for the user than rendering first:

"Profile language: English [edit]"

And then

"Profile language: [English ˆ]"
[Save] [Cancel]

Someday when there is more Additional information than just language dropdown, edit mode could be used. 

Changing a primary / removing an item does not require edit mode either.

src/profile/helpers/createActionAriaLabels.ts
```diff
@@ -0,0 +1,42 @@
...
+    const translationKey = getDataTypeAsTranslationKey();
+    const editTranslation = `profileInformation.ariaEdit${translationKey}`;
+    const primaryTranslation = `profileInformation.primary${translationKey}AriaDescription`;
+    const ariaSetPrimaryWithValue = `profileInformation.ariaSetPrimary${translationKey}WithValue`;
+    const ariaRemoveWithValue = `profileInformation.ariaRemove${translationKey}WithValue`;
+    return {
+      edit: `${t(editTranslation, { value })}`,
+      primary: `${t(primaryTranslation)}`,
+      setPrimary: `${t(ariaSetPrimaryWithValue, { value })}`,
+      remove: `${t(ariaRemoveWithValue, { value })}`,
```
akikoskinen: Do the `t` calls have to be within a string? Does `t` return something else than a string?

src/i18n/fi.json
```diff
@@ -138,7 +138,7 @@
     "postOffice": "Postitoimipaikka",
     "primary": "Ensisijainen",
     "remove": "Poista",
-    "setPrimary": "Aseta ensisijaiseksi",
+    "setPrimary": "Aseta ensisijaiseksi ",
```
akikoskinen: What's with the extra space?

src/i18n/fi.json
```diff
@@ -184,6 +184,12 @@
     "ariaListTitlePhones": "Antamasi puhelinnumerot",
     "ariaListTitleEmails": "Antamasi sähköpostiosoitteet",
     "ariaListTitleAddresses": "Antamasi osoitteet",
+    "ariaSetPrimaryAddressWithValue": "Aseta {{value}} ensisijaiseksi osoitteeksi",
+    "ariaSetPrimaryEmailWithValue": "Aseta {{value}} ensisijaiseksi sähköpostiksi",
+    "ariaSetPrimaryPhoneWithValue": "Aseta {{value}} ensisijaiseksi puhelinnumeroksi",
+    "ariaRemoveAddressWithValue": "Poista osoite {{value}} ",
+    "ariaRemoveEmailWithValue": "Poista sähköposti {{value}}",
```
akikoskinen: Pitäisikö käyttää pitempää "sähköpostiosoite" termiä, niin kuin näyttäisi, että on muissakin kohdissa käytetty?

src/i18n/en.json
```diff
@@ -184,6 +184,12 @@
     "ariaListTitlePhones": "EN: Antamasi puhelinnumerot",
     "ariaListTitleEmails": "EN: Antamasi sähköpostiosoitteet",
     "ariaListTitleAddresses": "EN: Antamasi osoitteet",
+    "ariaSetPrimaryAddressWithValue": "Set {{value}} as primary address",
+    "ariaSetPrimaryEmailWithValue": "Set {{value}} as primary email",
+    "ariaSetPrimaryPhoneWithValue": "Set {{value}} as primary phone",
+    "ariaRemoveAddressWithValue": "Remove address {{value}}",
+    "ariaRemoveEmailWithValue": "Remove email {{value}}",
+    "ariaRemovePhoneWithValue": "Remove phone {{value}}",
```
akikoskinen: Should the term "phone number" be used instead of "phone"? Both terms seem to be in use, but feels like "phone number" better refers to the _number_, whereas just "phone" refers to the device.

akikoskinen (commented):
A few more comments, nothing biggy.

src/profile/components/additionalInformation/AdditionalInformation.tsx
akikoskinen: > Changing a primary / removing an item does not require edit mode either.

Very good point, somehow I didn't realize this earlier. This is good now.

src/profile/helpers/createActionAriaLabels.ts
NikoHelle: No, t returns a string. Fixed. 
Even if a translation is not found for a key, it still returns the key.

src/i18n/fi.json
NikoHelle: From the translation Excel. Removed it

src/i18n/fi.json
NikoHelle: Joo, laitoin saman.

src/i18n/en.json
NikoHelle: Both seems to be used. I changed new and old ones to use "phone number"

akikoskinen (approved):
:+1:

PR #169, opened 2021-06-08
HP-318 add back Verified personal information

NikoHelle:
The branch merged to develop did not use VPI component. Copied it from the previous branch used in 'dev' and 'test'-servers.

Also removed 'subscriptions' from App.tsx. It was brought back in merge and previously removed in HP-548.


akikoskinen (changes requested):
Few improvement suggestions for commit messages, nothing else.

akikoskinen (approved):

PR #171, opened 2021-06-16
HP-727: HDS Update

NikoHelle:
Updated HDS to latest (1.1.0.) Small changes except with new Dropdowns and Dialog.

Body-element uses base HDS font-size and line-height. All "px" font sizes are replaced with HDS font sizes.

A lot of files are changed because all CSS classnames were renamed from 'className' to 'class-name'. This is the  coding standard defined in dev.hel.fi


src/i18n/en.json
```diff
@@ -122,7 +122,7 @@
     "cancel": "Cancel",
     "city": "City",
     "country": "Country",
-    "delete": "Delete",
+    "delete": "profileForm┊delete",
```
akikoskinen: What is this change? Is this translation actually used any where?

akikoskinen (commented):
Looks nice.

I did notice that the dialog that is shown when the user is about to delete the profile, has focus on the title and it draws a thick border around it.
modal-title-focus

Another observation, probably not related to this PR however: no matter what UI language the user selects, the header always says "Profile", and page title (the one within the <title></title> tags in HTML) is always "Profiili".

NikoHelle:
> Looks nice.
> 
> I did notice that the dialog that is shown when the user is about to delete the profile, has focus on the title and it draws a thick border around it.
> modal-title-focus
> 

That happens every time a dialog is opened. Also in  [HDS examples](https://hds.hel.fi/storybook/react/?path=/story/components-dialog--default) Focus is set for better accessibility. The focus border could be prettier, IMHO :). The heading is created by the HDS component, so I'm not overriding it at the moment.

> Another observation, probably not related to this PR however: no matter what UI language the user selects, the header always says "Profile", and page title (the one within the <title></title> tags in HTML) is always "Profiili".

Good point. I've added translations in a new commit.

src/i18n/en.json
NikoHelle: That's odd. I reverted that change.

akikoskinen:
> > I did notice that the dialog that is shown when the user is about to delete the profile, has focus on the title and it draws a thick border around it.
> 
> That happens every time a dialog is opened. Also in [HDS examples](https://hds.hel.fi/storybook/react/?path=/story/components-dialog--default) Focus is set for better accessibility. The focus border could be prettier, IMHO :). The heading is created by the HDS component, so I'm not overriding it at the moment.

Yep, no point trying to circumvent HDS features. But could it be communicated to HDS devs that this part could maybe use some improvement?

NikoHelle:
> > > I did notice that the dialog that is shown when the user is about to delete the profile, has focus on the title and it draws a thick border around it.
> > 
> > 
> > That happens every time a dialog is opened. Also in [HDS examples](https://hds.hel.fi/storybook/react/?path=/story/components-dialog--default) Focus is set for better accessibility. The focus border could be prettier, IMHO :). The heading is created by the HDS component, so I'm not overriding it at the moment.
> 
> Yep, no point trying to circumvent HDS features. But could it be communicated to HDS devs that this part could maybe use some improvement?

Has been done: [https://helsinkicity.slack.com/archives/CHCV3KTHA/p1623933531040900](https://helsinkicity.slack.com/archives/CHCV3KTHA/p1623933531040900)

akikoskinen (approved):

PR #172, opened 2021-06-23
HP-652: accessibility audit fixes

NikoHelle:
This PR fixes issues listed in the audit report. The related Jira-ticket lists the issues. The report itself is not included in the ticket, because it has sensitive information.

Fixed also some issues not listed, but reported by Axe, Wave, or jsx/a11y - or just needed improvement like better translations.

Two new components in commits
- (HP-652) List all form errors in an alert …
- (HP-652) Added aria helpers for inputs …
were in the branch used in the audit and added in this PR with other accessibility features.


src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
```diff
@@ -120,12 +121,8 @@ function VerifiedPersonalInformation(): React.ReactElement | null {
...
             ? t('profileInformation.verifiedDataInformationAriaPrefix')
             : t('profileInformation.verifiedDataInformationPrefix'),
-          // eslint-disable-next-line jsx-a11y/anchor-has-content
-          <a
-            href={t('profileInformation.verifiedDataInformationLink')}
-            target="_blank"
-            rel="noopener noreferrer"
-            tabIndex={forAria ? 0 : -1}
+          <NewWindowLinkWithChildrenAsTitle
+            link={t('profileInformation.verifiedDataInformationLink')}
```
akikoskinen, line 125: The resulting aria-label (in Finnish) is now:

> Ulkoinen sivusto: Tämä linkki avaa uuteen ikkunaan sivun "siirry suomi.fi palveluun "

While probably somewhat understandable, it does sound quite cumbersome. But looking at it more thoroughly, I think the cumbersomeness comes from the original text, which is even visible to those who can see:

> Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: [siirry suomi.fi palveluun](#)

IMO that's cumbersome text and needlessly describes how a link works (siirry...). A more fluent version could be:

> Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi [suomi.fi](#) verkkopalvelussa

With this change the aria-label would become:

> Ulkoinen sivusto: Tämä linkki avaa uuteen ikkunaan sivun "suomi.fi"

Much more fluent in my opinion.

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
```diff
@@ -32,13 +33,12 @@ function ProfileInformationAccountManagementLink(): React.ReactElement | null {
         />
       </div>
       <div className={styles['link']}>
-        <a
-          href={authenticationMethodReferenceUrl}
-          target="_blank"
-          rel="noopener noreferrer"
+        <NewWindowLink
+          link={authenticationMethodReferenceUrl}
+          title={t('profileInformation.doGoToAccountManagement')}
```
akikoskinen, line 38: Here's a similar issue as with the verified personal information long description text. The aria-label in Finnish is:

> Ulkoinen sivusto: Tämä linkki avaa uuteen ikkunaan sivun "Siirry tunnuksen hallintaan".

I think the problem here is also that the original visible text (Siirry tunnuksen hallintaan) isn't the best possible. I tried to think of a better one, but couldn't come up with one quickly.

But I think this link is also confusing to those who _can_ see (after these accessibility fixes, those who can't see are probably better informed). The link in no way suggests that it points to an external service. It could be improved. But I'm going on a tangential here, out of the scope of this PR. Just something to think about.

src/profile/components/accessibleFormikErrors/AccessibleFormikErrors.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [currentErrorText, setLastError] = useState<string>('');
+  const lastSubmitCountRef = useRef<number>(formik.submitCount);
+  const { t } = useTranslation();
+  const errorList = Object.keys(formik.errors);
+  const errorCount = errorList.length;
+  const formFields = getFormFields(props.dataType);
+  const errorsToText = () =>
+    errorList
+      .map(errorKey => {
+        const errorMessage = _.get(formik.errors, errorKey);
```
akikoskinen: Is `lodash` really needed here? Isn't this a simple property fetching from `object`?

src/common/accessibilityFieldHelpers/AccessibilityFieldHelpers.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+function createElementData({ dataType }: Props, t: TFunction): ElementData[] {
+  const validationRequiredTranslation = t('validation.required');
+  const dataFields = getFormFields(dataType);
+  return Object.keys(dataFields).map(
+    (key: string): ElementData => {
+      const fieldData = dataFields[key];
+      const fieldTranslation = t(fieldData.translationKey);
+      const translationKey = dataFields[key].comboBox
```
akikoskinen: `dataFields[key]` is already available as `fieldData`.

akikoskinen (commented):
Submitting comments thus far. Have to restart computer :smile:

src/common/pageMeta/PageMeta.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const { title } = props;
+  const { i18n } = useTranslation();
+  const scrollRef = useRef<boolean>(true);
+
+  useEffect(() => {
+    if (scrollRef.current) {
+      window.scrollTo(0, 0);
+      scrollRef.current = false;
+    }
+  }, [scrollRef]);
```
akikoskinen: What does this code achieve? The commit message gives no clue about the purpose.

src/accessibilityStatement/AccessibilityStatement.tsx
```diff
@@ -31,9 +32,10 @@ function AccessibilityStatement(): React.ReactElement {
   return (
     <div className={styles['page-wrapper']}>
       <Header />
-      <div className={styles['container']}>
+      <PageMeta title={t('footer.accessibility')} />
```
akikoskinen: This should probably have its own title in the translations, and not reuse a translation for another use.

akikoskinen (commented):
Went through the rest of the commits now.

src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
NikoHelle: The translation "Ulkoinen sivusto: Tämä linkki avaa uuteen ikkunaan sivun {(title}}" is new and requires fine-tuning of translations in some cases. It certainly is cumbersome with some of the current ones. The component accepts second prop "ariaLabelTitle" that can be used in cases where the visible link text does not fit in to the "Ulkoinen linkki"-template.

**Anyway, I'm getting emails about a Link -component** that is under development in HDS and is intended for the same scenario:
https://github.com/City-of-Helsinki/helsinki-design-system/pull/483

It is less verbose with the aria-label. I'll copy its features as it has approved accessibility: "SiteImprove tested accessibility". This component should be replaced with that one when it is released.

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
NikoHelle: Yes, the translations need better formatting when a link will open in a new window. 

https://helsinkisolutionoffice.atlassian.net/browse/HP-837

src/profile/components/accessibleFormikErrors/AccessibleFormikErrors.tsx
NikoHelle: formik.errors might not exist, so formik.errors[erroKey] would throw. But on the other hand in that case, the loop is not run and function is not called. So it can be formik.errors[erroKey]

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
NikoHelle: > I'm not sure what requirement is the basis for this commit. The form errors are now listed in a separate element, with `role="alert"`.
> 
> Already previously when a form field failed validation, the DOM was changed to something like this (shortened to contain only relevant parts):
> 
> ```
> <div>
>     <input id="basic-data-firstName" type="text" name="firstName" aria-invalid="true" aria-labelledby="basic-data-firstName-helper" value="" aria-describedby="basic-data-firstName-error">
> </div>
> <div id="basic-data-firstName-error">Tämä kenttä on pakollinen.</div>
> ```
> 
> So, the `input` element got an attribute `aria-invalid="true"` and also `aria-describedby="basic-data-firstName-error"`, which points to the `div` element containing the error message.
> 
> Was there some problem with this already existing way to communicate errors? Are the errors communicated now in two separate ways and screen reader users are flooded with duplicate information?

The error alert is triggered when the form is submitted. Without it the screen reader would indicate nothing is wrong and why the form did not submit. Aria-labels and aria-invalid are for single inputs, the alert for the whole form. 

This feature was in the audited branch and added to this branch with other accessibility features.

NikoHelle:
> Not sure where this commit originates. I can't find any mention in the accessibility report that the logo and title needs to be a link. But the real problem here IMO is that this implementation doesn't turn the logo and title in to a real link. Instead, they just surprisingly react to a mouse click. Perhaps use the `titleUrl` property of `Navigation` component instead?

When the user navigates to "/accessibility", there is no return link unless user is logged in. Or if the login fails, there is no link to anywhere. This change was made as a universal fallback and logo is usually a link to frontpage.

The logo + title are wrapped in `<a>` by the HDS-component. There is no href, but if 'titleUrl' was used, then the link would reload whole page. If 'onTitleClick' and 'titleUrl' were both used, it would still reload the page, because the component does not return the event object and event.preventDefault() cannot be called to prevent reload. 

Footer logo link is for the same purpose.

src/common/pageMeta/PageMeta.tsx
NikoHelle: When user navigates from page to page, the scroll position does not change. Usually not a problem, because navigation is done from the header. But, for example, in the mobile view after clicking "Saavutettavuusseloste", the content changes, but is shown midway or worse.

src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
akikoskinen: That looks convenient. It also has that "arrow out of a box" icon, which is an indicator about the nature of the link even for those with normal vision.

src/common/pageMeta/PageMeta.tsx
akikoskinen: That's what I thought it would do. But when scrolling a bit down, and clicking links, the position didn't change. That's with desktop Firefox. Any idea what's going on?

src/profile/components/verifiedPersonalInformation/VerifiedPersonalInformation.tsx
NikoHelle: Copied the aria-label and icon from the Link PR. Created a ticket for future update: https://helsinkisolutionoffice.atlassian.net/browse/HP-838

HDS estimated the component will be released in August.

src/profile/components/accessibleFormikErrors/AccessibleFormikErrors.tsx
NikoHelle: Fixed.

src/common/accessibilityFieldHelpers/AccessibilityFieldHelpers.tsx
NikoHelle: Done.

src/common/pageMeta/PageMeta.tsx
NikoHelle: Found it. Not a Firefox issue. The component did not fully re-render on subpage change, so the useEffect was not run. Now it runs on every "title" change.

src/accessibilityStatement/AccessibilityStatement.tsx
NikoHelle: Added a new translation

src/accessibilityStatement/AccessibilityStatementEn.tsx
```diff
@@ -38,8 +40,11 @@ function AccessibilityStatementEn(): ReactElement {
         accessible format, they can request for this information by e-mail at{' '}
         <a href="mailto:helsinki.palaute@hel.fi">helsinki.palaute@hel.fi</a> or
         through the feedback form at{' '}
-        <a href="https://www.hel.fi/feedback">www.hel.fi/feedback</a>. The aim
-        is to reply to the enquiry within a reasonable time frame.{' '}
+        <NewWindowLink
+          link="https://www.hel.fi/feedback"
+          title="www.hel.fi/feedback"
+        />
+        . The aim is to reply to the enquiry within a reasonable time frame.{' '}
```
akikoskinen: The extra space in the end is probably not necessary.

src/common/newWindowLink/NewWindowLink.tsx
```diff
@@ -0,0 +1,53 @@
...
+    children,
+    className,
+    ...rest
+  } = props;
+  const { t } = useTranslation();
+  const getAriaLabel = () => {
+    const text = t('opensInNewWindow', {
+      title: `${ariaLabelTitle || title}`,
+    });
+    return text.slice(-1) === '.' ? text.substr(0, text.length - 1) : text;
```
akikoskinen: This removes a full stop from the end of string, right? All the `opensInNewWindow` translations have a full stop in the end. Two questions:
1. why mustn't there be a full stop in the end?
2. why is the full stop removed in code, and not simply in the translations themselves?

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
```diff
@@ -32,13 +33,13 @@ function ProfileInformationAccountManagementLink(): React.ReactElement | null {
...
+          hideIcon
```
akikoskinen, line 39: I think this link would be more understandable to those with proper vision, if the icon _is_ shown.

NikoHelle: There's the ">" icon from already, that's why the second one is hidden. The "new window"-icon does not fit the layout.

src/common/newWindowLink/NewWindowLink.tsx
NikoHelle: Quite often the link text/title does not include a full stop - especially when links are created by code. Like with NewWindowLinkWithChildrenAsTitle.

1. there might be two of them and really can't force all translations to include one.
2. The "." is in the translation file to ensure the sentences are separated: "{{title}}. Opens in new window." And new translations are added probably without knowing what code does, so better to have the "." in the template

src/accessibilityStatement/AccessibilityStatementEn.tsx
NikoHelle: Yeah, doesn't seem necessary as other translations do not have it.

akikoskinen:
> > Not sure where this commit originates. I can't find any mention in the accessibility report that the logo and title needs to be a link. But the real problem here IMO is that this implementation doesn't turn the logo and title in to a real link. Instead, they just surprisingly react to a mouse click. Perhaps use the `titleUrl` property of `Navigation` component instead?
> 
> When the user navigates to "/accessibility", there is no return link unless user is logged in. Or if the login fails, there is no link to anywhere. This change was made as a universal fallback and logo is usually a link to frontpage.

The "tabs" are still there, so the user can navigate to "Omat tiedot" (which _is_ the "front page"). Or they can use the browser's back button of course. But I'm fine with having the header logo as a link to front page as that really is customary nowadays. That brings us back to the _real_ problem...

> The logo + title are wrapped in `<a>` by the HDS-component. There is no href, but if 'titleUrl' was used, then the link would reload whole page. If 'onTitleClick' and 'titleUrl' were both used, it would still reload the page, because the component does not return the event object and event.preventDefault() cannot be called to prevent reload.

Is this a problem with the HDS's `Navigation` component? Are you saying that it can't be configured to create an `<a>` element with `href` attribute in a way that the whole page wouldn't be reloaded? If so, can a feature request be sent to HDS to make this possible? (And while waiting for that feature, I'm fine with this current solution, if there's no better way now)

> Footer logo link is for the same purpose.

And there the link really is a proper link :+1:.

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
akikoskinen: Does the ">" icon have a common meaning? I haven't figured one out. To me it indicates that it could be an accordion, or a menu or whatever. _Instead of_ the ">" icon, the "arrow out of box" icon would more clearly indicate an external link.

src/common/pageMeta/PageMeta.tsx
akikoskinen: Works now. But not sure if this is a good feature _for every link_. For example, scroll just a bit down, click another tab (or e.g. change UI language) -> position jumps to top. Feels very weird.

NikoHelle:
> > > Not sure where this commit originates. I can't find any mention in the accessibility report that the logo and title needs to be a link. But the real problem here IMO is that this implementation doesn't turn the logo and title in to a real link. Instead, they just surprisingly react to a mouse click. Perhaps use the `titleUrl` property of `Navigation` component instead?
> > 
> > 
> > When the user navigates to "/accessibility", there is no return link unless user is logged in. Or if the login fails, there is no link to anywhere. This change was made as a universal fallback and logo is usually a link to frontpage.
> 
> The "tabs" are still there, so the user can navigate to "Omat tiedot" (which _is_ the "front page"). Or they can use the browser's back button of course. But I'm fine with having the header logo as a link to front page as that really is customary nowadays. That brings us back to the _real_ problem...
> 
> > The logo + title are wrapped in `<a>` by the HDS-component. There is no href, but if 'titleUrl' was used, then the link would reload whole page. If 'onTitleClick' and 'titleUrl' were both used, it would still reload the page, because the component does not return the event object and event.preventDefault() cannot be called to prevent reload.
> 
> Is this a problem with the HDS's `Navigation` component? Are you saying that it can't be configured to create an `<a>` element with `href` attribute in a way that the whole page wouldn't be reloaded? If so, can a feature request be sent to HDS to make this possible? (And while waiting for that feature, I'm fine with this current solution, if there's no better way now)

"Normal" links do not work in SPA:s, because clicking them loads a new page - like it should. To prevent this, the click event's preventDefault() should be called. But the HDS component do not pass this event to the onClick-listener. I'd consider this as a bug and will report this to the HDS developers.

Side note: SPAs usually have an UI-library that provides some kind of Link-component. It prevents click events and handles browser history. Profile UI uses the common React Router for this, so there aren't many "event.preventDefault()" -lines in the code.

> 
> > Footer logo link is for the same purpose.
> 
> And there the link really is a proper link 👍.

src/common/pageMeta/PageMeta.tsx
akikoskinen: In the accessibility statement link's case this could be a useful feature, but perhaps not for every link.

src/common/newWindowLink/NewWindowLink.tsx
akikoskinen: I don't get it. This doesn't remove a duplicate full stop. It turns e.g.

```
Privacy policy. Opens in a new window.
```
into
```
Privacy policy. Opens in a new window
```
removing the final full stop from the string.

The string is also used as an `aria-label` (only), so does it make a difference if there is a full stop in the end or not?

akikoskinen:
> [--] the click event's preventDefault() should be called. But the HDS component do not pass this event to the onClick-listener. I'd consider this as a bug and will report this to the HDS developers.

Very good. Hopefully it'll get fixed. Until that, we'll come by with the current situation.

src/common/pageMeta/PageMeta.tsx
NikoHelle: But in non-SPA pages, the jump would happen anyway when you navigate from page to page? The feature in not link related, it happens when page changes.

src/accessibilityStatement/AccessibilityStatementEn.tsx
NikoHelle: Fixed

src/common/newWindowLink/NewWindowLink.tsx
NikoHelle: No wonder, it was hard to understand - there was a bug and I was blind to it. The "." should be removed from the **title**, not from the resulting text. Sorry! Fixed now.

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
NikoHelle: In the design, the ">" emphazised the text "Go to..." That's how I understand it. And the text+icon is aligned to the right because the ">" points that direction. There would be no need for that alignment, if there was only "Jump + box" :)

src/common/pageMeta/PageMeta.tsx
akikoskinen: Hmm, that seems to be true. For example this Github PR page has a similar feature: when one clicks on the tabs, the position jumps to the top. I hadn't really noticed that before :astonished: Perhaps it's quite ok then.

src/common/newWindowLink/NewWindowLink.tsx
akikoskinen: Should there be some test(s) for this component?

src/profile/components/profileInformation/ProfileInformationAccountManagementLink.tsx
akikoskinen: Have any designs actually used this "arrow out of box" icon for external links? Consistency is also important in UX and I guess it would be good to use only one solution to indicate any given thing.

But yeah, we're going perhaps a bit too much on the territory of UI design here, where the topic was actually to improve accessibility. We can leave the ">" icon be as it is.

src/common/newWindowLink/NewWindowLink.tsx
NikoHelle: Added!

src/common/newWindowLink/__tests__/NewWindowLink.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  it('removes "." from end of title so aria-label do not have double ("..")', async () => {
+    const title = 'test with dot. And another title.';
+    const element = renderComponentAndReturnElement({
+      title,
+      link,
+    });
+    expect(element.getAttribute(ariaLabelAttribute)).toEqual(
+      `${title.substr(0, title.length - 1)}${labelTranslation}`
+    );
```
akikoskinen: In order to not use the same algorithm in production and test code _and_ to make the idea of the test more explicit, I would suggest a small change to this test.

```suggestion
    const titleWithoutDot = 'the title';
    const titleWithDot = `${titleWithoutDot}.`;
    const element = renderComponentAndReturnElement({
      title: titleWithDot,
      link,
    });
    expect(element.getAttribute(ariaLabelAttribute)).toEqual(
      `${titleWithoutDot}${labelTranslation}`
    );
```

akikoskinen (commented):
Nice tests added. One small suggestion.

akikoskinen (approved):

PR #176, opened 2021-06-24
HP-834: Upgrade react-scripts

NikoHelle:


akikoskinen (commented):
This update seems to bring in a new version of pretty much every babel plugin and some other packages, _in addition_ to the already existing version. For example there are now no less than three different versions of `@babel/preset-env` in the `yarn.lock` file (there were already two before this PR). Is there a way to avoid this massive increase in different package versions?

NikoHelle:
React-scripts is a dependency monster. But this is probably a yarn issue - same with lodash-versions in another PR. 
Have to run clean yarn install for this one too, after https://github.com/City-of-Helsinki/open-city-profile-ui/pull/179

NikoHelle:
After clean install, this PR looks better too.

akikoskinen (approved):
Yes, much cleaner now.

PR #177, opened 2021-06-24
HP-622: Updated graphQL and removed subscriptions related data

NikoHelle:
Ran codegen which updates generatedTypes.ts.
Then fixed typings.ts accordingly.


akikoskinen (approved):
:+1:

PR #179, opened 2021-06-28
Re-installed package.json dependencies and cleaned yarn lock file

NikoHelle:
Yarn did not seem to dedupe dependencies correctly and left multiple versions of packages that have mutual version requirements. Removed yarn.lock (and node_modules) and ran "yarn".

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,10 +2,10 @@
 # yarn lockfile v1
 
 
-"@ahooksjs/use-request@^2.8.4":
-  version "2.8.4"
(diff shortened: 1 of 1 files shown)


akikoskinen (approved):
Quite a many updated packages at the same time. But if testing seems to show that this is all good, then why not.

NikoHelle:
Odd number of updates, I'd say. But most of the packages are for react-scripts etc and UI uses only a few.  

Tested with yarn lint, yarn test, yarn build, yarn start and serve build. No hiccups or errors shown.

PR #182, opened 2021-08-03
HP-791: Error page

NikoHelle:
Created a dedicated error page and route which renders it. A new useErrorPageRedirect -hook is used for redirecting user to the ErrorPage with optional title, message and props to hide buttons.

Note: there is no UI design for this page.

Error page parameters can tested by just going to the error page url: 
http://localhost:3000/error?title=new%20title&message=new%20message&hideLoginButton=1&hideFrontPageButton=1


src/profile/components/errorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+import authService from '../../../auth/authService';
+
+export type ErrorPageQueryParams = {
+  message?: string;
+  title?: string;
+  hideLoginButton?: string;
+  hideFrontPageButton?: string;
+};
+
+export const errorPagePath = '/error';
```
akikoskinen: I think the error path shouldn't be in this component. The error path sounds more like a setting. Is there a common place for settings?

Also this same setting should also be used in the App's route specification.

src/profile/components/errorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+              {t('login.login')}
+            </Button>
+          )}
+          {hideFrontPageButton === undefined && (
+            <Button
+              onClick={() => history.push('/')}
+              data-testid={'error-page-frontpage-button'}
+            >
+              {t('nav.goToHomePage')}
+            </Button>
```
akikoskinen: To my eye it would be better to have the "go to home page" as a link, not as a button.

src/profile/components/errorPage/__tests__/ErrorPage.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+      - a title
+      - a message 
+      - hideLoginButton with a value 
+      - hideFrontPageButton with a value`, async () => {
+    const title = 'test_title';
+    const message = 'test_message';
+    const hideLoginButton = 'anyValue';
+    const hideFrontPageButton = 'anyValue';
+    // eslint-disable-next-line max-len
+    mockUseLocationValue.search = `title=${title}&message=${message}&hideLoginButton=${hideLoginButton}&hideFrontPageButton=${hideFrontPageButton}`;
```
akikoskinen: Why not split the long line to multiple lines, if linter complains about it?

src/profile/components/errorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+    title,
+    hideLoginButton,
+    hideFrontPageButton,
+  } = queryParamsToObject<ErrorPageQueryParams>(location.search);
+  const notificationMessage = message || t('notification.defaultErrorText');
+  const notificationTitle = title || t('notification.defaultErrorTitle');
+  const isAuthenticated = authService.isAuthenticated();
+
+  return (
+    <PageLayout title={title} data-testid="error-page-layout">
```
akikoskinen: Should the `PageLayout`'s title also be set to `notificationTitle`, in order to use the default title when needed?

akikoskinen (commented):
Some comments, nothing major.

src/profile/components/errorPage/ErrorPage.tsx
NikoHelle: There is config, so I placed the path there. And also used it in App.tsx

src/profile/components/errorPage/ErrorPage.tsx
NikoHelle: True, buttons should be used for actions not for links. Moved the link after the message.

src/profile/components/errorPage/__tests__/ErrorPage.test.tsx
NikoHelle: Because the same prop is repeated in all tests,  it is quite long before this and splitting didn't improve readability much.
It is splitted now! :)

src/profile/components/errorPage/ErrorPage.tsx
NikoHelle: Thx! Changed this.

src/profile/components/errorPage/__tests__/ErrorPage.test.tsx
akikoskinen: Just wondering, is there a way to split a string into many lines in JavaScript? Something like in C++:

```c++
const auto myStr = "very long "
    "string";
```

or Python:

```python
my_str = ("very long "
    "string")
```

akikoskinen (approved):

src/profile/components/errorPage/__tests__/ErrorPage.test.tsx
NikoHelle: It is possible with template strings:

```
mockUseLocationValue.search = `title=${title}\
&message=${message}\
&hideLoginButton=${hideLoginButton}\
&hideFrontPageButton=${hideFrontPageButton}`;
```

The "\" at the end of line, tells js to continue string without a line break. But then cannot have indentation (space/tab) on each line because those would be in the string. And the code would look weird when few lines are not indented at all.

akikoskinen: Got it. Would string catenation also work?

```javascript
const myStr = "very long " +
    "string";
```

NikoHelle: Yes, that style works too.

PR #186, opened 2021-08-23
HP-896: add new env var for GDPR auth token

NikoHelle:
Added the variable and fixed how it is read from env variables: not from process.env, but from window._env_. Like all other env variables.

PROFILE_AUDIENCE cannot be used as client id in the gdprAuthCodeManager, because changing the PROFILE_AUDIENCE also affected OIDC_SCOPE. PROFILE_AUDIENCE is also used as an object key in pickProfileApiToken. So new variable was needed.

Current set-up worked in DEV and TEST, because gdprAuthCodeManager read PROFILE_AUDIENCE from process.env and not from window._env_.


akikoskinen (approved):
Looks Ok to me.

PR #187, opened 2021-08-24
HP-836: optimize graphql queries

NikoHelle:
When users navigate to their profile page, the page loads
- ProfileExistsQuery
- MyProfileQuery
and also
- ServiceConnectionsQuery
- GdprServiceConnectionsQuery 

Neither ServiceConnectionQuery is required when profile is shown. Loading started, because DeleteProfile ja DownloadProfile components used useQuery-hooks instead of useLazyQuery. 

DeleteProfile requires ServiceConnections to be loaded so the UI can show correct information before deleting the profile. Had to add a loader and load indicator. Now the UI loads the connections first and then shows proper warnings. 

DownloadProfile won't show any additional data to user, so lazy query was easy to add there.

### DIFF
--- a/src/profile/components/deleteProfile/DeleteProfile.tsx
+++ b/src/profile/components/deleteProfile/DeleteProfile.tsx
@@ -1,9 +1,9 @@
-import React, { useEffect, useState } from 'react';
-import { useQuery } from '@apollo/client';
+import React, { useCallback, useEffect, useState } from 'react';
+import { useLazyQuery } from '@apollo/client';
 import { loader } from 'graphql.macro';
 import { useTranslation } from 'react-i18next';
 import * as Sentry from '@sentry/browser';
-import { Button, Checkbox } from 'hds-react';
+import { Button, Checkbox, LoadingSpinner } from 'hds-react';
 import { useHistory } from 'react-router';
 import { useMatomo } from '@datapunt/matomo-tracker-react';
 
@@ -25,6 +25,14 @@ const SERVICE_CONNECTIONS = loader(
 function DeleteProfile(): React.ReactElement {
   const [deleteConfirmationModal, setDeleteConfirmationModal] = useState(false);
   const [deleteInstructions, setDeleteInstructions] = useState(false);
+  const notStartedLoadState = 'not-started';
+  const loadingLoadState = 'loading';
+  const loadedLoadState = 'loaded';
+  const [dataLoadState, setDataLoadState] = useState<
+    | typeof notStartedLoadState
+    | typeof loadingLoadState
+    | typeof loadedLoadState
+  >(notStartedLoadState);
   const { createToast } = useToast();
   const history = useHistory();
   const { trackEvent } = useMatomo();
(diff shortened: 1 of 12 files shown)


src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -52,24 +60,52 @@ function DeleteProfile(): React.ReactElement {
...
+    }
+  }, [getServiceConnections, setDataLoadState, dataLoadState]);
+
+  const onExpandingPanelChange = useCallback(
+    isOpen => {
+      if (isOpen) {
+        loadServiceConnections();
+      }
+    },
+    [loadServiceConnections]
```
akikoskinen, line 106: Could the service connections fetching be delayed even more, to the point when the last warning modal is shown?

akikoskinen:
> When users navigate to their profile page, the page loads
> 
>     * ProfileExistsQuery
>     * MyProfileQuery

Not sure if it's part of this PR, but to me it seems that the `ProfileExistsQuery` is needless. It uses the `myProfile` GraphQL query to check if a profile exists. Then, if profile exists, it executes the `MyProfileQuery`, which **again** uses `myProfile` GraphQL query to fetch the profile data. It should be enough to use the `myProfile` GraphQL query only once.

src/profile/components/deleteProfile/DeleteProfile.tsx
NikoHelle: Good idea! Fetching **could** be done after user clicks the delete button! 

I tried it... and looks like there is a major accessibility issues there, if the load animation is in the ConfirmationModal. The ConfirmationModal is "accessibility-wise" an aria alert. If the content of the modal is loaded later and changes, it's hard to indicate to screen readers what happened. For example buttons appear after loading. 

Also UX-wise it would seem that pressing the delete button would start the deletion (as the user sees something is loading), but it would only load the ServiceConnections and then ask "Are you sure".  Seems a bit odd and if it takes a long time, user doesn't know was the profile deleted now or not.

Also there are some issues with the Dialog component, if the heading changes while it is open. The heading is auto-focused when Dialog opens and loses focus when heading element changes. Having two different modals would solve this, but is it worth it...?

Too many issues rises, if the fetching is moved, so I'd leave it there. 😄

NikoHelle:
> > When users navigate to their profile page, the page loads
> > ```
> > * ProfileExistsQuery
> > * MyProfileQuery
> > ```
> 
> Not sure if it's part of this PR, but to me it seems that the `ProfileExistsQuery` is needless. It uses the `myProfile` GraphQL query to check if a profile exists. Then, if profile exists, it executes the `MyProfileQuery`, which **again** uses `myProfile` GraphQL query to fetch the profile data. It should be enough to use the `myProfile` GraphQL query only once.

Thx, this should be changed too! At the moment the ProfileContext cannot handle this scenario. It only loads, keeps and mutates complete profile data. But it cannot handle the scenario, where it profile does not exist. This raise an error. And it cannot reload the profile after is has been created. 

So, I'll make a ticket for this and change it later.

NikoHelle:
FYI: I added a new commit regarding double loading of the api tokens.

src/profile/components/deleteProfile/DeleteProfile.tsx
akikoskinen: Good thoughts. Makes sense to leave the loading where it is now. But could the `t('deleteProfile.explanation')` text be shown already immediately, and the spinner would roll below it? It's quite a long text so the user could already read it while the loading happens.

NikoHelle: Good idea. Done.

akikoskinen (approved):
:+1:

PR #190, opened 2021-09-02
HP-884: only one email in the UI

NikoHelle:
User can add or edit only one email. That email is the primary email.

Changed quite many tests, because emails cannot be added (except one), removed or set as primary in the UI. Some component renaming too because of that.


src/profile/helpers/editData.ts
```diff
@@ -550,7 +559,16 @@ export function createEditorForDataType(
...
+      if (
+        dataType === 'emails' &&
+        allItems.findIndex(item => item.primary) > -1
+      ) {
+        throw new Error('Cannot add a new email if a primary already exists');
+      }
+      const newNode = createNewProfileNode(
+        dataType,
+        dataType === 'emails' ? { primary: true } : undefined
+      );
```
akikoskinen, line 571: As the email data model is still similar to the addresses and phones (i.e. there can be many of them), is it the job of this data handling code to limit emails to only one and force it to be primary? Could that be a UI side restriction only?

akikoskinen (commented):
One architectural question. Otherwise looks good to me.

src/profile/helpers/editData.ts
NikoHelle: UI is meant to be a "stupid" viewing component and all data related restrictions and creation etc. is in editData.ts.

That "`dataType === 'emails' ? { primary: true } : undefined`" had to be added, because UI cannot edit the data it receives directly. The data it is read only. UI can just make calls to the editData to mutate the data.

There is a function "setPrimary" that works with phone numbers and addresses (and did work with multiple emails), but that makes an API call.

So I preferred this approach where editData still has full control of the data and creates a new email with primary=true.

akikoskinen: Got it.

akikoskinen:
Rebase so that conflicts get resolved. Then I think this is good to go.

akikoskinen (approved):

PR #191, opened 2021-09-14
HP-879 session polling

NikoHelle:
Same polling system that was added first to Example UI.

Same files:
- http-poller
- __mocks__/http-poller

Polling is controller by authService, which does not exist in Example U, so it's new code. Polling is bound to events triggered by the oidc-client. It is also started when UI is loaded and a valid user exists in session storage.

Testing can be done by starting Profile UI locally and logging in to Profile UI (dev) or Example UI (dev) with same user. And then logout from dev UI.  Change poll interval to 5000ms or wait 60 secs for polling to check user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "hds-core": "1.1.0",
     "hds-design-tokens": "1.1.0",
     "hds-react": "1.1.0",
+    "http-status-typed": "^1.0.0",
     "i18n-iso-countries": "^6.8.0",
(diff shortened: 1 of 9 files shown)


akikoskinen (commented):
Http poller seems to be missing tests (those are present in example UI).

NikoHelle:
> Http poller seems to be missing tests (those are present in example UI).

Oh, yes! I copied them too.

akikoskinen (approved):

PR #192, opened 2021-09-17
HP-941: Unify translations used in develop and KuVa prod

NikoHelle:
Develop version and KuVa prod version had incompatible translation files. The "one email only"-feature in development added and removed translations, but removed translations are needed in Kuva prod. 
 
The translation excel was simply reverted to the version used in Kuva prod and translations used in develop where added.

Meanwhile SV and EN versions where translated, so sv.json and en.json show a lot more changes than fi.json.


src/i18n/en.json
```diff
@@ -168,31 +171,38 @@
...
     "primaryPhone": "Primary phone number",
     "primaryAddressAriaDescription": "This is your primary address",
+    "primaryEmailAriaDescription": "This is your primary email",
     "primaryPhoneAriaDescription": "This is your primary phone number",
-    "verifiedDataInformationPrefix": "EN: Merkityt",
-    "verifiedDataInformationAriaPrefix": "EN: Alla olevat",
-    "verifiedDataInformation": "EN: <0></0> tiedot on vahvistettu virallisesta väestötietojärjestelmästä, eikä niitä ei voi tämän takia muokata Helsinki Profiilissa. Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: <1>siirry suomi.fi palveluun </1>",
+    "verifiedDataInformationPrefix": "Marked",
+    "verifiedDataInformationAriaPrefix": "Those below",
+    "verifiedDataInformation": "<0></0> the data has been verified from the official Population Information System and they can thus not be edited in the Helsinki Profile. You can check your data in the Population Information System through the suomi.fi web service: <1>go to the suomi.fi service </1>",
```
akikoskinen, line 184: Isn't it so that the `<0></0>` in `verifiedDataInformation` is replaced with either `verifiedDataInformationPrefix` or `verifiedDataInformationAriaPrefix`? Doing those replacements doesn't produce very good sentences.

It's not a good idea to build sentences in this way at all, as it will produce bad results in some language(s). There should probably be different translations for the visible case and the aria case, even if they would contain mostly the same words.

src/i18n/en.json
```diff
@@ -168,31 +171,38 @@
...
+    "verifiedBasicData": "Verified basic data",
     "phones": "Phone numbers",
+    "emails": "Emails",
     "addresses": "Addresses",
     "ariaComboBoxHelperText": "Select {{field}}. Search by typing and select one option. {{requiredInfo}}",
     "ariaInputFieldHelperText": "Write {{field}}.  {{requiredInfo}}",
-    "ariaListTitlePhones": "EN: Antamasi puhelinnumerot",
-    "ariaListTitleAddresses": "EN: Antamasi osoitteet",
+    "ariaListTitlePhones": "Phone numbers that you have provided",
+    "ariaListTitleEmails": "Email adresses that you have provided",
```
akikoskinen: Typo: "adresses" should be "addresses".

akikoskinen (changes requested):

src/i18n/en.json
NikoHelle: Yes, this is ugly and quite unreadable. In this case React components are inserted in to the translations:
```
<Trans
        i18nKey="profileInformation.verifiedDataInformation"
        components={[
          forAria
            ? t('profileInformation.verifiedDataInformationAriaPrefix')
            : t('profileInformation.verifiedDataInformationPrefix'),
          <NewWindowLinkWithChildrenAsTitle
            link={t('profileInformation.verifiedDataInformationLink')}
            tabIndex={forAria ? 0 : -1}
          />,
        ]}
      />
```
Components are inserted by index, that's why there is <0>, <1> This is the way i18next have [documented](https://react.i18next.com/latest/trans-component) the usage.

Looks like there is also newer way of using keys instead of indexes. This usage was copied from the usage of another translation (profileForm.terms) which dates before keys could be used.

NikoHelle: This PR unifies the translation files, so same files can be used in master and develop branches. Switching from indexes to keys, would require code changes in both branches, cherry-picked of course. But not in this PR, perhaps a new ticket for this?

akikoskinen: > not in this PR, perhaps a new ticket for this?

That works.

src/i18n/en.json
NikoHelle: Good catch. Fixed in excel + translations.

src/i18n/en.json
akikoskinen: In this case the same result can be achieved with both the indices and the keys. The main problem here is probably that the translator didn't have enough context how the translated strings are used. So this is not that much a technical problem but a communication problem.

There are (at least) two options to solve that:
1. communicate more/better
2. reduce the need for communication, i.e. make the translator's job easier

I think point 2 should be tried first and only if that isn't sufficient, then use point 1.

Here's a quick sketch of a diff what the translation file could look like so that it doesn't have to communicate that much context to the translator:

```diff
-"verifiedDataInformationPrefix": "EN: Merkityt",
-"verifiedDataInformationAriaPrefix": "EN: Alla olevat",
-"verifiedDataInformation": "EN: <0></0> tiedot on vahvistettu virallisesta väestötietojärjestelmästä, eikä niitä ei voi tämän takia muokata Helsinki Profiilissa. Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: <1>siirry suomi.fi palveluun </1>",
+"verifiedDataInformation": "EN: Merkityt tiedot on vahvistettu virallisesta väestötietojärjestelmästä, eikä niitä ei voi tämän takia muokata Helsinki Profiilissa. Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: <0>siirry suomi.fi palveluun </0>",
+"verifiedDataInformationAria": "EN: Alla olevat tiedot on vahvistettu virallisesta väestötietojärjestelmästä, eikä niitä ei voi tämän takia muokata Helsinki Profiilissa. Voit tutustua halutessasi väestötietojärjestelmään tallennettuihin tietoihisi suomi.fi verkkopalvelussa: <0>siirry suomi.fi palveluun </0>",
```

akikoskinen (approved):

src/i18n/en.json
NikoHelle: I agree, if translation process needs better communication, there is a bigger problem beneath. :)

PR #200, opened 2021-10-05
HP-975: 404 page not found

NikoHelle:
Modified the ErrorPage component, so it can be used as a 404 error page too.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,6 +20,7 @@ import GdprAuthorizationCodeManagerCallback from './gdprApi/GdprAuthorizationCod
 import ToastProvider from './toast/ToastProvider';
 import authService from './auth/authService';
 import config from './config';
+import PageNotFound from './common/pageNotFound/PageNotFound';
 
 countries.registerLocale(fi);
 countries.registerLocale(en);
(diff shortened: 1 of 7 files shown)


src/common/page404/Page404.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import ErrorPage from '../../profile/components/errorPage/ErrorPage';
+
+function Page404(): React.ReactElement {
```
akikoskinen: I wouldn't use the text "404" in this component's name, because this component **does not** make the server return a 404 HTTP status code, when a not found page (or route) is requested. Perhaps call this component `PageNotFound` or something like that.

akikoskinen (changes requested):

src/common/page404/Page404.tsx
NikoHelle: "404Page" was first my thought, but a function name cannot start with a number. Sounds weird that naming it "Page404" would suggest that a React component returns http status 404. 

Renamed it to PageNotFound.

NikoHelle:
Removed the part of the commit message that was redundant.

src/common/page404/Page404.tsx
akikoskinen: 404 is just a number, without any inherent meaning. If interpreted as an HTTP status code, it means "Not Found". In other contexts it can mean something else or be meaningless. The feature in this PR has nothing to do with HTTP or HTTP status codes. So in this context the number 404, at least to me, means nothing (or then I can't figure out what the context here would be where 404 has some meaning).

The 404 is not in code any more, but is still in the commit message. Could remove it from there too.

NikoHelle: Commit message changed.

akikoskinen (approved):

PR #201, opened 2021-10-05
HP-1011: Fix default language code bug

NikoHelle:
In some browsers the default language returned by i18n was "en-GB" or "fi-FI" which would also be seen in UI.


akikoskinen (commented):
Would be good to write a test for this, if possible.

NikoHelle:
Tests added

akikoskinen (approved):

PR #202, opened 2021-10-05
HP-992: Update design of the Create profile -page

NikoHelle:
Current CSS is not very responsive. Copied styles from ViewProfile -page:
- koros at the top
- same bg-color
- same spacings

There is a ticket HP-1012 for unifying paddings and margins. That is why those are not unified in this ticket.


akikoskinen (approved):
Nice

PR #199, opened 2021-10-04
HP-975: Update footer

NikoHelle:
Imported the footer component from latest version of HDS. Its background colors do not match hel.fi, so had to override it.

The login page was dark blue, so it looked bad with a black footer, so Login page was re-designed. Therefore this PR has changes to Login page too. 

Colors and Login page changes are base on this design: https://app.abstract.com/projects/71b36afc-ac4f-4b35-8bf2-5a5e255c72c6/branches/master/commits/69109acf2c6172fbf43090d1157e3df0f1019629/files/c8a30a32-3a2b-4a9c-8146-96800a169517/layers/3315F227-3D39-4EF0-9381-8F5D1757EDB0?mode=build&selected=root-DD6900BD-50DB-4BBA-8DB0-E08C57B3CAF2&sha=69109acf2c6172fbf43090d1157e3df0f1019629

There is a ticket HP-1012 for unifying paddings and margins. That is why those are not unified in this ticket.

PR #204, opened 2021-10-06
HP-1001: Fix error handling in profile api fetch

NikoHelle:
Error handling was incorrect when api returned any other error than graphQL error

API does not return an error when a profile does not exist, just an empty myProfile.

So if an error is returned, it is always fatal whether it is a graphQL or a network error.

Added better notfication for the user and buttons for retry and logout.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -111,7 +111,9 @@
     "authenticating": "Authenticating..."
   },
   "profile": {
-    "loading": "Loading profile..."
+    "loading": "Loading profile...",
+    "loadErrorText": "EN: Profiilin haku epäonnistui! Yritä hakea uudelleen tai kirjaudu ulos.",
+    "reload": "Retry"
   },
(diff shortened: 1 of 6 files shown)


akikoskinen:
Some discoveries. Not all of these are necessarily part of this PR.

- The header bar dropdown menus (e.g. language selection) go behind the error notification.
- When the new error notification is shown, page title still says "Create profile".
- The buttons below the error notification have no spacing to the footer. Perhaps this gets fixed with #199 or some other PR?
- The error text could be a tad shorter. Perhaps it includes a bit more details than is necessary.

NikoHelle:
I rebased to develop -branch, which has never version of HDS. Now the Notification is not above menus
Also the error text has been cut and now the page title has error title.
The new footer and updated HDS did not improve padding, so added more.

akikoskinen (approved):
Looks good.

PR #208, opened 2021-10-14
HP-974: HDS font-sizes and colors

NikoHelle:
Use responsive heading font-sizes from HDS and remove "own" font sizes, weights and colors. Also changed some css class names that were not very informative (main, small).

Removed also background color override from footer.

Body text size is correct, so did not change those.


akikoskinen (approved):
Deleting code is good :+1:

PR #209, opened 2021-10-21
HP-1027: code cleanup

NikoHelle:
Removed some unused components ja fixed some styles.

HDS has now a Link-component which is almost identical to NewWindowLink. It lacks some functionalities that are needed, so copied the HDS link and added the missing functionalities.


src/common/copyOfHDSLink/Link.tsx
```diff
@@ -0,0 +1,181 @@
...
+   */
+  hideIcon?: boolean;
+  /**
+   * Element placed on the left side of the link text
+   */
+  iconReplacement?: React.ElementType;
+  /**
+   * Element placed on the left side of the link text
+   */
+  'aria-label'?: string;
```
akikoskinen: The documentation here is incorrect.

src/i18n/en.json
```diff
@@ -237,8 +237,42 @@
...
   "opensInNewWindow": "{{title}}. Opens in a new window.",
+  "openInNewTabAriaLabel": "Opens in a new tab.",
+  "openInExternalDomainAriaLabel": "EN: Siirtyy toiseen sivustoon.",
   "accessibilityStatement": "Accessibility Statement",
   "cityOfHelsinki": "City of Helsinki",
   "pageNotFoundTitle": "Page not found",
-  "pageNotFoundText": "EN: Tarkista URL-osoitteen oikeinkirjoitus (tarkista isot ja pienet kirjaimet sekä välimerkit). Tai palaa edelliselle sivulle napsauttamalla selaimen Edellinen-painiketta. Voit mennä myös etusivulle alla olevasta linkistä."
-}
+  "pageNotFoundText": "EN: Tarkista URL-osoitteen oikeinkirjoitus (tarkista isot ja pienet kirjaimet sekä välimerkit). Tai palaa edelliselle sivulle napsauttamalla selaimen Edellinen-painiketta. Voit mennä myös etusivulle alla olevasta linkistä.",
+  "cookies": {
```
akikoskinen: The cookie translations probably shouldn't be included in this PR.

src/accessibilityStatement/AccessibilityStatementSv.tsx
```diff
@@ -46,10 +46,13 @@ function AccessibilityStatementSv(): React.ReactElement {
...
-          link="https://www.hel.fi/helsinki/sv/stad-och-forvaltning/delta/feedback"
-          title="www.hel.fi/feedback"
-        />
+        <Link
+          href="https://www.hel.fi/helsinki/sv/stad-och-forvaltning/delta/feedback"
+          external
+          openInNewTab
+        >
+          www.hel.fi/feedback
+        </Link>
```
akikoskinen, line 55: This problem wasn't introduced by this PR, but the href and visible URL don't match.

NikoHelle: Probably the "wrong" text is used, because the url is so looooooong it doesn't fix mobile views:

image

It doesn't really matter where the user is (re)directed? Also in the English version, the user is finally redirected elsewhere than the page shows.

src/i18n/en.json
NikoHelle: Fixed!

src/common/copyOfHDSLink/Link.tsx
NikoHelle: Changed!

src/accessibilityStatement/AccessibilityStatementSv.tsx
akikoskinen: Perhaps the URL doesn't even have to be included in the text. But this topic doesn't need to be addressed in this PR anyway.

NikoHelle: Yes, I wondered about it too and I guess it's there because the link before it (helsinki.palaute@hel.fi) also has the "content" visible? Copywriting issue :)

src/common/copyOfHDSLink/Link.tsx
```diff
@@ -0,0 +1,181 @@
...
+   */
+  hideIcon?: boolean;
+  /**
+   * Element placed on the left side of the link text
+   */
+  iconReplacement?: React.ElementType;
+  /**
+   * The computed aria-label can be overridden with this prop
+   */
+  'aria-label'?: string;
```
akikoskinen: It looks weird that this one property uses kebab-case, while the others use camelCase. Perhaps stick to the convention.

NikoHelle: The "aria-label" is a HTML-attribute, and "ariaLabel" would need to be converted back to "aria-label" when setting it to the element. Now it can be set by using just "{...rest}" in line 150.

But anyway "aria-label" is already included in the element props, because it inherits from "a"-element. So I just removed it from the file.

akikoskinen (approved):

PR #212, opened 2021-10-21
HP-839: correct service connection creation time

NikoHelle:
List of connected services showed the creation time of the service. It should show creation time of the service connection.


src/profile/graphql/ServiceConnectionsQuery.graphql
```diff
@@ -18,6 +18,7 @@ query ServiceConnectionsQuery {
               }
             }
           }
+          createdAt
```
akikoskinen, line 20: Is the `Service`'s `createdAt` used for anything any more? Could it be removed from the fetched fields in the GraphQL call?

NikoHelle: Good point. Removed it.

akikoskinen (approved):

PR #216, opened 2021-11-04
HP-1057: Fixed service connection load bug

NikoHelle:
Apollo-client does not call 'onCompleted' on cache hit.

The data object returned from useLazyQuery-hook is now also checked to detect, if loading is complete due a cache hit.

Checked also other onCompleted-callbacks. They all have 'no-cache' or 'network-only'. Except profile download, but caching cannot happen, because downloading causes a full page reload.

### DIFF
--- a/src/common/expandingPanel/ExpandingPanel.tsx
+++ b/src/common/expandingPanel/ExpandingPanel.tsx
@@ -16,6 +16,7 @@ type Props = PropsWithChildren<{
   initiallyOpen: boolean;
   scrollIntoViewOnMount?: boolean;
   onChange?: (isOpen: boolean) => void;
+  dataTestId?: string;
 }>;
 
 function ExpandingPanel({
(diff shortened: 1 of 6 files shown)


akikoskinen:
Would there be a test for this?

NikoHelle:
There has never been unit tests for DeleteProfile or similar component DownloadProfile. Probably because the whole flow is  quite complex to test. It would require to mock service connections, authorisation codes and deletion results.

I'll add them. Takes few days.

NikoHelle:
Tests added. The "When re-rendered, service connections are fetched from cache..." -test will fail without the changes in the initial commit "68a700d".

I noticed an UX bug and fixed that also in this PR, because it is was found while creating tests.

akikoskinen (approved):

akikoskinen (approved):

PR #217, opened 2021-11-04
HP-1049: generic profile deletion errors

NikoHelle:
When profile deletion failed, a berth error was checked. That error does not exist anymore.

Converted berth error handling in to a generic error handling.


src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -45,12 +46,12 @@ function DeleteProfile(): React.ReactElement {
...
-        setBerthError(true);
-      } else {
-        Sentry.captureException(error);
-        createToast({ type: 'error' });
-      }
+      const standardErrorMessage =
+        error.graphQLErrors && error.graphQLErrors.length
+          ? error.graphQLErrors[0].message
+          : error.message;
+      Sentry.captureException(new Error(standardErrorMessage));
```
akikoskinen: This sends only the first GraphQL error to Sentry. Why not send all of them, if there are many?

src/profile/components/modals/deleteProfileError/DeleteProfileError.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+  const getErrorTranslationKey = (resultError: Props['error']): string => {
+    if (!resultError) {
+      return 'deleteProfileErrorModal.genericError';
+    }
+    const graphQlErrors = (resultError as ApolloError).graphQLErrors;
+    if (graphQlErrors) {
+      const filteredErrors: string[] = graphQlErrors
+        .map(graphQlError => graphQlError?.extensions?.code)
+        .filter(graphQlError => graphQlError);
```
akikoskinen: Naming thoughts:

- `filteredErrors` could be named `errorCodes`, because that's what it seems to contain.
- `map` "extracts" the `code` from the errors, so the filter line could be: `.filter(code => code)`

src/profile/components/deleteProfile/DeleteProfile.tsx
NikoHelle: Can there be many, if possible errors are:

- PROFILE_DOES_NOT_EXIST_ERROR
- MISSING_GDPR_API_TOKEN_ERROR
- CONNECTED_SERVICE_DELETION_NOT_ALLOWED_ERROR
- CONNECTED_SERVICE_DELETION_FAILED_ERROR

There is no specification for Sentry, so can't say what should be sent. Previous version did not sent graphQL errors at all.

But the DeleteProfileError.tsx assumes there can be multiple, so this should too.

src/profile/components/modals/deleteProfileError/DeleteProfileError.tsx
NikoHelle: Fixed. I removed the array.filter as it is not necessary. array.includes works without it.

src/profile/components/deleteProfile/DeleteProfile.tsx
akikoskinen: I think it always needs to be assumed that there can be multiple errors. Also the documented error code lists aren't necessarily exhaustive.

akikoskinen (approved):

PR #219, opened 2021-11-08
HP-1064: Updated HDS and removed copy of Notification component

NikoHelle:
Old HDS had version of react-spring that caused errors in built package.


akikoskinen (approved):
Looks simple enough.

PR #222, opened 2021-11-24
HP-1099: Render the verify email notification only with certain amr

NikoHelle:
First commit includes missing tests for useProfile. Tests were added because second commit includes a change how useProfile handles the profile returned from Tunnistamo. According to the OIDC spec, Profile.amr should be an array of strings, but it is not. Now it is converted to an array, if it is not.

EmailEditor should show the "verify your email"-notification only if user's amr is 'heltunnistussuomifi'. Currently that is the only case, so the check is simple. In other cases the backend won't send an email and the notification should not be shown.

### DIFF
--- a/src/auth/__tests__/useProfile.test.tsx
+++ b/src/auth/__tests__/useProfile.test.tsx
@@ -1,0 +1,124 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+import { User } from 'oidc-client';
+
+import {
+  mockUserCreator,
+  MockedUserOverrides,
+} from '../../common/test/userMocking';
+import authService from '../authService';
+import useProfile, { Profile } from '../useProfile';
+
+type Status = 'loading' | 'error' | 'loaded';
+type DataGetters = {
+  getInfo: () => Status;
+  getProfile: () => Profile | null;
+  hasCalledGetUser: () => boolean;
+  getMockedUserData: () => User;
+};
+
+describe('useProfile', () => {
+  const loadingStatus: Status = 'loading';
+  const loadedStatus: Status = 'loaded';
+  const errorStatus: Status = 'error';
+  const statusIndicatorElementId = 'status-indicator';
+  const profileElementId = 'profile';
+  const noProfile = { noProfile: true };
+  const TestProfileComponent = ({
+    callCounter,
+  }: {
+    callCounter: () => number;
+  }) => {
+    const hasLoadStarted = callCounter() > 0;
+    const { profile, loading, error } = useProfile();
+    const isFinished = hasLoadStarted && loading === false;
+
+    if (error) {
+      return <span id={statusIndicatorElementId}>{errorStatus}</span>;
+    }
+    if (!isFinished) {
+      return <span id={statusIndicatorElementId}>{loadingStatus}</span>;
+    }
+    return (
+      <div>
+        <span id={statusIndicatorElementId}>{loadedStatus}</span>
+        <span id={profileElementId}>
+          {JSON.stringify(profile ? profile : noProfile)}
+        </span>
+      </div>
+    );
+  };
+  const renderTestComponent = (
+    overrides?: MockedUserOverrides,
+    error = false
+  ): DataGetters => {
+    const userData = mockUserCreator(overrides);
+    const mockedGetUser = jest.spyOn(authService, 'getUser');
+
+    if (error) {
+      mockedGetUser.mockRejectedValueOnce(null);
+    } else {
+      mockedGetUser.mockResolvedValueOnce(userData);
+    }
+    const result = render(
+      <TestProfileComponent
+        callCounter={() => mockedGetUser.mock.calls.length}
+      />
+    );
+    const { container } = result;
+    const getElementById = (id: string) =>
+      container.querySelector(`#${id}`) as HTMLElement;
+    return {
+      getInfo: () =>
+        getElementById(statusIndicatorElementId).innerHTML as Status,
+      getProfile: () => {
+        const data = getElementById(profileElementId).innerHTML;
+        return JSON.parse(data);
+      },
+      hasCalledGetUser: () => mockedGetUser.mock.calls.length > 0,
+      getMockedUserData: () => userData,
+    };
+  };
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should return the profile which the authService.getUser() provides', async () => {
+    const {
+      getInfo,
+      getProfile,
+      getMockedUserData,
+      hasCalledGetUser,
+    } = renderTestComponent();
+    const userData = getMockedUserData();
+    await waitFor(() => expect(getInfo()).toEqual(loadedStatus));
+    expect(hasCalledGetUser()).toBeTruthy();
+    expect(getProfile()).toEqual(userData.profile);
+  });
+
+  it('should provide no profile if it has expired', async () => {
+    const { getInfo, getProfile } = renderTestComponent({
+      userOverrides: ({
+        expired: true,
+      } as unknown) as Partial<User>,
+    });
+    await waitFor(() => expect(getInfo()).toEqual(loadedStatus));
+    expect(getProfile()).toEqual(noProfile);
+  });
+
+  it('should provide no profile if user.expired is undefined', async () => {
+    const { getInfo, getProfile } = renderTestComponent({
+      userOverrides: ({
+        expired: undefined,
+      } as unknown) as Partial<User>,
+    });
+    await waitFor(() => expect(getInfo()).toEqual(loadedStatus));
+    expect(getProfile()).toEqual(noProfile);
+  });
+
+  it('should return error when authService.getUser() fails', async () => {
+    const { getInfo } = renderTestComponent(undefined, true);
+    await waitFor(() => expect(getInfo()).toEqual(errorStatus));
+  });
+});
--- a/src/auth/__tests__/useProfile.test.tsx
+++ b/src/auth/__tests__/useProfile.test.tsx
@@ -84,17 +84,24 @@ describe('useProfile', () => {
     jest.restoreAllMocks();
   });
 
-  it('should return the profile which the authService.getUser() provides', async () => {
+  it('should return the profile which the authService.getUser() provides where amr is always an array', async () => {
+    const amr = 'string-arm';
     const {
       getInfo,
       getProfile,
       getMockedUserData,
       hasCalledGetUser,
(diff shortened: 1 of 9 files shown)


src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
```diff
@@ -52,3 +52,7 @@ export function getAmrUrl(authenticationMethodReference: AMRStatic): string {
       );
   }
 }
+
+export function hasTunnistusSuomiFiAmr(profile: Profile | null): boolean {
+  return profile?.amr === tunnistusSuomifiAMR;
```
akikoskinen: According to [the OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html#IDToken), the `amr` claim is an **array of strings**. It could be that Tunnistamo erroneously provides a single string. I think both styles should be handled here, a single string, and an array of strings.

akikoskinen (changes requested):
Should the `amr` values (plural, as I think there could be more than one of them) that trigger showing the notification, be configurable? I suppose that the set of those values can change and also be specific to different environments.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
NikoHelle: Very good point, I didn't know Tunnistamo is not compliant with the specs. I checked that Helsinki Tunnistus (Keycloak) returns an array in amr. But Profiili does not use Tunnistus (yet).

Tunnistamo never returns an array, so the possibility should not be added to Profiili at the moment. The code would be redundant. It is clear this must be changed when Tunnistamo won't be used anymore.

NikoHelle:
> Should the `amr` values (plural, as I think there could be more than one of them) that trigger showing the notification, be configurable? I suppose that the set of those values can change and also be specific to different environments.

I added just this simple check, because it is unclear what kind of scenarios there will be in the future. According to Slack-conversations nobody knows exactly how the email verification will evolve and when those are sent in the future.

So until there is a second scenario, which might show the way, I'd just stick to this simple way. If the email sending depends only on amr and that changes in different environments, then it's definitely a good idea to add it to env config.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
akikoskinen: That would be knowingly leaving a bug in the code waiting to bite. I wouldn't leave it there. Who remembers to fix it on time before it bites?

NikoHelle: From Profile's point of view the bug is in Tunnistamo. :)

I'll change this to be future proof.

Maybe report a bug for Tunnistamo?

akikoskinen: > ...the bug is in Tunnistamo. :)
> Maybe report a bug for Tunnistamo?

Yep, that's known. Unfortunately it's difficult to change because it would break backward compatibility :disappointed:

NikoHelle: Tests added and amr is now converted to "string[]".

src/auth/__tests__/useProfile.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const { getInfo, getProfile } = renderTestComponent({
+      userOverrides: ({
+        expired: true,
+      } as unknown) as Partial<User>,
+    });
+    await waitFor(() => expect(getInfo()).toEqual(loadedStatus));
+    expect(getProfile()).toEqual(noProfile);
+  });
+
+  it('should return an empty array is arm is undefined', async () => {
```
akikoskinen: First "is" should probably be "if".

src/auth/__tests__/useProfile.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    await waitFor(() => expect(getInfo()).toEqual(loadedStatus));
+    expect(hasCalledGetUser()).toBeTruthy();
+    const profileWithConvertedAmr = {
+      ...userData.profile,
+      amr: [amr],
+    };
+    expect(getProfile()).toEqual(profileWithConvertedAmr);
+  });
+
+  it('should return null if user.expired !== false', async () => {
```
akikoskinen: This test name could be more readable. Perhaps something like:
> should provide no profile if it has expired

src/profile/components/emailEditor/__tests__/EmailEditor.test.tsx
```diff
@@ -27,6 +27,19 @@ import { EditDataType, EmailValue } from '../../../helpers/editData';
 import i18n from '../../../../common/test/testi18nInit';
 import RenderChildrenWhenDataIsComplete from '../../../../common/test/RenderChildrenWhenDataIsComplete';
 import getEmailsFromNode from '../../../helpers/getEmailsFromNode';
+import { mockProfileCreator } from '../../../../common/test/userMocking';
+import { AMRStatic } from '../../../../auth/useProfile';
+
+type TestAmr = Extract<AMRStatic, 'google'> | 'foobar';
```
akikoskinen: Instead of "foobar", could the test amr be something more descriptive, like perhaps "tunnistus-suomi-fi"?

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
```diff
@@ -52,3 +56,7 @@ export function getAmrUrl(authenticationMethodReference: AMRStatic): string {
       );
   }
 }
+
+export function hasTunnistusSuomiFiAmr(profile: Profile | null): boolean {
+  return extractSingleAmr(profile) === tunnistusSuomifiAMR;
```
akikoskinen: This should check if `tunnistusSuomifiAMR` is present _anywhere_ in the `profile`'s `amr` array.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
```diff
@@ -5,8 +5,12 @@ import {
 } from '../../../auth/useProfile';
 import config from '../../../config';
 
+function extractSingleAmr(profile: Profile | null): string {
+  return profile ? profile.amr[0] : '';
```
akikoskinen: What if the `amr` array is empty? Should a check be made before executing `profile.amr[0]`?

akikoskinen (commented):
Some smallish comments.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
NikoHelle: When does the amr array has length !== 1? The spec says it is possible, but in our environments?

From the amr spec: "Parties using this claim will need to agree upon the meanings of the values used, which may be context-specific." Is there an agreement somewhere?

UI does not process more than one amr value. For example the ProfileInformation component renders only one ProfileInformationAccountManagementLink. If there are no amr values, the component is not rendered so user won't notice it.

src/profile/components/emailEditor/__tests__/EmailEditor.test.tsx
NikoHelle: This is intentionally silly value so it is not mistaken to the actual value.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
NikoHelle: This depends on the conversation above.

src/profile/components/emailEditor/__tests__/EmailEditor.test.tsx
akikoskinen: It could still have some meaning. Perhaps something like "fake-suomi-fi-amr-for-test"...

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
akikoskinen: Currently I guess there's only one `amr` value. I'm not aware of any case where there would be more than one value in the list. I just thought that finding the correct value from anywhere in the array would require about the same amount of code, so why not do it.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
akikoskinen: This is also relevant if the `amr` claim can be missing or an empty string (in the case of Tunnistamo, which doesn't provide the amrs in an array but as a string).

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
NikoHelle: Number of items in the "amr" array affects other places too. Now it can be assumed there is 0 or 1 item and the value can just be picked. Otherwise all checks should search if the value exists anywhere in the array.

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
NikoHelle: The string value is converted to an array of single string. But anyway the return value of that function requires "| undefined"

src/profile/components/profileInformation/profileInformationAccountManagementLinkUtils.ts
akikoskinen: Got it. Since that's the case, I guess the multi-value support can be left for another time.

src/profile/components/emailEditor/__tests__/EmailEditor.test.tsx
NikoHelle: Switched to a value defined in "AmrStatic", because it is most convenient. Switching to a value longer than 10 chars will trigger linting error "sonarjs/no-duplicate-string"

src/auth/__tests__/useProfile.test.tsx
NikoHelle: Done!

src/auth/__tests__/useProfile.test.tsx
NikoHelle: The returned value of user.expired can be "undefined", if the expires_in the token is "undefined".  That is why "user.expired !== false" is used instead of just "!user.expired". 

But the name of the test does not need to reflect that. Added another test for undefined.

akikoskinen (approved):

PR #223, opened 2021-12-09
HP-894: Apollo client update

NikoHelle:
The update caused few minor issues:
- "headers.Authorization" changed to "headers.authorization"
- reference to graphQL error objects changes between renders. This caused the useEffect-hook with dependency to the error object to be called on each render
- React started to give warnings when a child component called ProfileContext while parent component was still rendering

### DIFF
--- a/src/graphql/client.ts
+++ b/src/graphql/client.ts
@@ -16,7 +16,7 @@ const authMiddleware = new ApolloLink((operation, forward) => {
     operation.setContext({
       headers: {
         authorization: `Bearer ${token}`,
-        'Accept-Language': i18n.languages[0],
+        'accept-language': i18n.languages[0],
       },
     });
   }
--- a/src/graphql/client.ts
+++ b/src/graphql/client.ts
@@ -15,7 +15,7 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   if (token) {
     operation.setContext({
       headers: {
-        Authorization: `Bearer ${token}`,
+        authorization: `Bearer ${token}`,
         'Accept-Language': i18n.languages[0],
       },
     });
--- a/src/profile/hooks/useProfileQuery.ts
+++ b/src/profile/hooks/useProfileQuery.ts
@@ -1,6 +1,6 @@
 import { ApolloError, FetchResult, useLazyQuery } from '@apollo/client';
 import { loader } from 'graphql.macro';
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 import { usePersistFn } from 'ahooks';
 import _ from 'lodash';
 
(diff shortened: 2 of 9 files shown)


src/graphql/client.ts
```diff
@@ -15,7 +15,7 @@ const authMiddleware = new ApolloLink((operation, forward) => {
   if (token) {
     operation.setContext({
       headers: {
-        Authorization: `Bearer ${token}`,
+        authorization: `Bearer ${token}`,
         'Accept-Language': i18n.languages[0],
```
akikoskinen: Should all headers, not just `authorization`, now be lower cased? For consistency.

yarn.lock
```diff
@@ -13156,6 +13157,11 @@ tslib@^2.0.0, tslib@^2.0.1, tslib@^2.0.3, tslib@^2.1.0, tslib@^2.2.0:
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.3.0.tgz#803b8cdab3e12ba581a4ca41c8839bbb0dacb09e"
   integrity sha512-N82ooyxVNm6h1riLCoyS9e3fuJ3AMG2zIZs2Gd1ATcSFjSA23Q0fzjjZeh0jbJvWVDZ0cJT8yaNNaaXHzueNjg==
 
+tslib@^2.3.0:
+  version "2.3.1"
```
akikoskinen: There is now `tslib` 2.3.0 and `tslib` 2.3.1 in the dependencies. One of those should be enough.

yarn.lock
```diff
@@ -6969,13 +6963,20 @@ graphql-tag@2.12.4:
...
-graphql-tag@^2.10.1, graphql-tag@^2.12.0:
+graphql-tag@^2.10.1:
   version "2.12.5"
   resolved "https://registry.yarnpkg.com/graphql-tag/-/graphql-tag-2.12.5.tgz#5cff974a67b417747d05c8d9f5f3cb4495d0db8f"
   integrity sha512-5xNhP4063d16Pz3HBtKprutsPrmHZi5IdUGOWRxA2B6VF7BIRGOHZ5WQvDmJXZuPcBg7rYwaFxvQYjqkSdR3TQ==
   dependencies:
     tslib "^2.1.0"
 
+graphql-tag@^2.12.3:
+  version "2.12.6"
```
akikoskinen: There is now `graphql-tag` 2.12.5 and `graphql-tag` 2.12.6 in the dependencies. One of those should be enough.

src/profile/hooks/useProfileQuery.ts
```diff
@@ -29,6 +29,10 @@ type QueryReturnType = {
 export function useProfileQuery(props?: {
   onError: (queryError: ApolloError) => void;
 }): QueryReturnType {
+  const errorToTriggerRef = useRef<ApolloError | undefined>(undefined);
```
akikoskinen, line 32: A few lines above this is a comment about [a bug in apollo-client](https://github.com/apollographql/apollo-client/issues/5531) that had to be circumvented. That bug is claimed to be fixed in Apollo client version 3.5.0. Can the previously non-usable functionality now be used?

akikoskinen (changes requested):

src/graphql/client.ts
NikoHelle: Yes, I changed that one too.

yarn.lock
NikoHelle: This is fixed now.  How come yarn cannot detect these 🙄

yarn.lock
NikoHelle: Fixed this too.

src/profile/hooks/useProfileQuery.ts
NikoHelle: I tried that, but there were complications:
- when `errorPolicy: 'all'` is used, the onError callback is not triggered, if there is a graphQL-error
- if errorPolicy has another value, it doesn't allow graphQL errors or those are totally ignored. Then the profile won't load, if user is logged in with weak authentication. In that case there is always a graphQL error present stating user don't have rights to load verified personal information.

So still have to use a this hack.  I'll change the comment about the bug

akikoskinen: OK. The comment wasn't yet changed in the changes currently present. Is it coming, or did you have second thoughts? Otherwise this is looking good.

NikoHelle: Just took some time to re-write and rebase :) Done now.

NikoHelle: Got code smells from trailing spaces in the comments. Updated again.

akikoskinen (approved):

PR #226, opened 2021-12-20
HP-1174:  authentication improvements

NikoHelle:
Minor changes to authService:
- clean up on sign out is focused to one place and depends more on triggered events
- improved error handling in api token fetching 
- when user token is silently renewed, api token fetching is retried if it fails. Profile has been fetched already once and there is time to refetch api tokens before they expire.
- when user is initially logged in and api token fetching fails, an error is shown (as before). In this case retrying is not necessary, because profile cannot be loaded and user has no data to lose.

### DIFF
--- a/src/auth/__tests__/authService.test.js
+++ b/src/auth/__tests__/authService.test.js
@@ -2,7 +2,10 @@ import to from 'await-to-js';
 import { waitFor } from '@testing-library/react';
 
 import authService, { API_TOKEN } from '../authService';
-import { getHttpPollerMockData } from '../__mocks__/http-poller';
+import {
+  getHttpPollerMockData,
+  enableActualHttpPoller,
+} from '../__mocks__/http-poller';
 import i18n from '../../common/test/testi18nInit';
 
 describe('authService', () => {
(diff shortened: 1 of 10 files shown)


src/auth/authService.ts
```diff
@@ -168,24 +210,49 @@ export class AuthService {
...
+  async fetchAndStoreApiToken(user: User): Promise<boolean> {
+    const [, response] = await to(this.fetchApiToken(user));
+    if (!response) {
+      return Promise.resolve(false);
+    }
+    const apiToken = await this.parseAndStoreApiToken(response);
+    return Promise.resolve(!!apiToken);
+  }
+
+  async cleanUpAfterUserRemovedOrSignedOut(): Promise<void> {
```
akikoskinen: Function names should represent what the function does. Not assume situations when it is called.

src/auth/http-poller-with-promises.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import createHttpPoller, { HttpPoller } from './http-poller';
+
+export type RetryingPollerProps = {
+  pollFunction: () => Promise<Response | undefined>;
+  delayInMs?: number;
+  maxRetries?: number;
+};
+
+export default async function retryPollingUntilSuccessful({
```
akikoskinen, line 11: To me this seems like a module that on the general level tries something again and again until successful. In this case it's tightly coupled to the HttpPoller interface, but it doesn't need to be. The general case has already been implemented, for example in [retry](https://www.npmjs.com/package/retry) or [p-retry](https://www.npmjs.com/package/p-retry). Could those be used instead of reinventing the wheel?

akikoskinen (commented):
A question for starters. I might have a few more smaller comments, but let's check this first.

src/auth/authService.ts
NikoHelle: Changed to cleanUpUserSessionAndApiTokens

src/auth/http-poller-with-promises.ts
NikoHelle: The idea is to couple this to the http-poller as it does what is needed. After adding the onSuccess -callback, of course. There wasn't much to reinvent as we already have a module that makes async polling. 

Adding a new module seemed like an overkill vs few new lines of code + tests.

src/auth/__tests__/authService.test.js
```diff
@@ -269,18 +283,87 @@ describe('authService', () => {
...
           "apiToken",
-          "71ffd52c-5985-46d3-b445-490554f4012a",
+          "${apiToken}",
         ]
       `);
     });
   });
 
+  describe(`Api tokens are fetched again after user tokens are renewed.
+            After silent renew completes, the _userLoaded event is raised and`, () => {
```
akikoskinen: The test name seems to end abruptly. I'm not sure should it continue more or is the final "and" unnecessary.

src/auth/__tests__/authService.test.js
```diff
@@ -269,18 +283,87 @@ describe('authService', () => {
...
+    beforeAll(() => {
+      enableActualHttpPoller(jest.requireActual('../http-poller'));
+    });
+
+    it('it calls fetchApiToken and stores new token after success', async () => {
+      const { sessionUser } = setSession({
+        validUser: true,
+        validApiToken: true,
+      });
+      sessionStorage.setItem(API_TOKEN, 'foobar');
```
akikoskinen: I always like to give test data such contents that it helps figuring out what's wrong if the test shall ever fail. For example here instead of "foobar" this could be "old api token" or something like that. If `apiToken` was equal to "new api token", then if the final assert in this test would fail, an error message would be printed that could look something like this (completely made up example, I don't know how the used test framework really prints errors):

```
Expected `sessionStorage.getItem(API_TOKEN)` to be equal to `apiToken`.
    `sessionStorage.getItem(API_TOKEN)` == "old api token"
    `apiToken` == "new api token"
```

src/auth/__tests__/authService.test.js
```diff
@@ -269,18 +283,87 @@ describe('authService', () => {
...
+
+      const fetcApiTokenSpy = jest.spyOn(authService, 'fetchApiToken');
+
+      await userManager.events._userLoaded.raise(sessionUser);
+      expect(fetcApiTokenSpy).toHaveBeenCalledTimes(1);
+      jest.advanceTimersByTime(501);
```
akikoskinen: Where is this magic number coming from (it's present more than once in this file)? I'm guessing there's probably a value 500 somewhere and this is supposed to be more than that. So perhaps that magic number 500 can be used here and then +1 to it. So something like: `theUltimateDelay + 1`.

src/auth/http-poller-with-promises.ts
```diff
@@ -0,0 +1,51 @@
...
+  pollFunction,
+  delayInMs = 1000,
```
akikoskinen: http-poller uses the term `pollIntervalInMs` for this. It could be named the same here.

src/auth/authService.ts
```diff
@@ -8,13 +8,33 @@ import {
...
+  }
+  const getExpiresIn = (expiresAtInSeconds?: number): number => {
+    if (expiresAtInSeconds) {
+      const now = Date.now() / 1000;
+      return expiresAtInSeconds - now;
+    }
+    return 0;
+  };
+
+  return getExpiresIn(user.expires_at) <= 0;
```
akikoskinen: I find this a bit hard to read. How about simply:

```suggestion
  if (user.expires_at) {
    return user.expires_at < Date.now() / 1000;
  }
  
  return true;
```

akikoskinen (changes requested):
Here comes the smaller comments.

src/auth/http-poller-with-promises.ts
akikoskinen: OK. That's fine.

src/auth/__tests__/authService.test.js
NikoHelle: The test group title is supposed to work as a prefix for the actual tests. The output of the tests would be combination of those. For example

describe(`Function xyz() ")
   it("returns true when arguments have something")
   it("returns false when arguments have something else")
   
Would output   
  
'Function xyz() returns true when arguments have something'
'Function xyz() returns false when arguments have something else'

I changed the titles a bit. Added "..." to indicate that text continues below.

src/auth/__tests__/authService.test.js
NikoHelle: This is a good point! Changed. 

The printed error shows assumed value and received value.

src/auth/__tests__/authService.test.js
NikoHelle: Added describing constant for this

src/auth/http-poller-with-promises.ts
NikoHelle: Changed!

src/auth/authService.ts
NikoHelle: Yeah, true. Previously the returned time was used in another context, but no need to return numbers anymore.

src/auth/__tests__/authService.test.js
akikoskinen: Ahh, of course. I missed that.

src/auth/authService.ts
akikoskinen: Now it's a bit more easier to follow the logic. But there still is the helper function that is called from only one place. While the name of the function provides some documentation, I think in this case it's not really necessary. Comparing some timestamp to the current time is pretty basic stuff.

NikoHelle: Removed the inner helper function and added tests for the isAuthenticatedUser() which uses this function.

akikoskinen (approved):

PR #227, opened 2021-12-21
HP-1169: Service connection texts do not update

NikoHelle:
Texts for the Service connections are fetched from the API. Those were updated only, if user was on the Service connections -tab when  language was changed.

Otherwise texts stored in the cache were used. Fixed the bug by clearing the cache when language changes.


NikoHelle:
BE was updated to accept language as variable, so this PR is useless.

PR #233, opened 2022-01-17
HP-1169 service connection query fix

NikoHelle:
Added current language as a variable when querying for service connections. When language is changed, the texts are refetched instead of cache hit, which caused the bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "open-city-profile-ui",
-  "version": "1.1.11",
+  "version": "1.1.12",
   "license": "MIT",
   "private": true,
   "dependencies": {
--- a/public/env-config.js
+++ b/public/env-config.js
@@ -14,5 +14,5 @@ window._env_ = {
   REACT_APP_PROFILE_BE_GDPR_CLIENT_ID: 'https://api.hel.fi/auth/helsinkiprofile',
   REACT_APP_OIDC_RESPONSE_TYPE: 'code',
   REACT_APP_SENTRY_DSN: 'https://8b5b23e2171b42cd8617e2b1ad7353b6@sentry.hel.ninja/63',
-  REACT_APP_VERSION: '1.1.11'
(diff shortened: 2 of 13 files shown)


akikoskinen:
I tried this and found an issue:

- Load Profile UI
- Don't change tabs, stay on the "front page"
- Switch language

Result: `ServiceConnections` fetch is executed but fails, because the `language` variable is not provided in the GraphQL query.

NikoHelle:
Thx, for the bug catch! Fixed...

akikoskinen:
Now there's another issue: the service connections are fetched twice at language change. First with the old language, and then immediately with the new language.

NikoHelle:
*face_palm

Ah, there is no need to track language changes as Apollo detects it on re-render anyway. Fixed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "open-city-profile-ui",
-  "version": "1.1.11",
+  "version": "1.1.12",
```
akikoskinen, line 3: Should version changes also be reflected in the `public/env-config.js` file's `REACT_APP_VERSION` property? At least for me `yarn start` always wants to consolidate those values.

NikoHelle: Yeah, env-config.js is re-created every time server starts, so I don't see why it is in the repo. The process how the file is created, was done by Kuva, so I don't have first hand info.

There is a ticket (https://helsinkisolutionoffice.atlassian.net/browse/HP-1184) which repairs a bug in the process. The env-config.js could be added to .gitignore in that one. 

I'll ask Kuva developers why it is not ignored yet.

akikoskinen: Ok. So still within this PR, should the version number be changed in env-config.js file as well?

NikoHelle: I asked Kuva. env-config is in .gitignore, but as it is also in the repo so not ignored. It can be removed.

But anyway I added the file here as it technically should be here until fixed.

akikoskinen (approved):

PR #234, opened 2022-01-19
HP-1184: "version" and "name" to env-config.js in Docker image

NikoHelle:
$npm_*** environment variables do not exist when env.sh is run. So "version" and "name" are empty in built Docker images.

Dockerfile now copies the package.json to the same location as env.sh, so it can read "version" and "name" -properties from the file and write them to env-config.js.

Deleted env-config.js from repo because it is re-created every server start.


akikoskinen:
Has this been tested to see that it works? To me it seems that the Docker image still gets the `.env` file from this repository and `env.sh` goes through it. The `.env` file still refers to `$npm_package_name` and `$npm_package_version` and this PR doesn't seem to change the fact that those envvars don't exist.

If I understood correctly, those envvars will exist when a command is run via the `package.json` scripts. So should some execution step be moved there, and run via `npm run ...` or something like that?

Of course it's completely possible that I have misunderstood how this all works.

NikoHelle:
.env has 
[key]: [value]

env.sh checks, if the key exists as an environmental variable. Values, like $npm_xxx, are ignored. This way also env vars from Azure library are set correctly. And this way $npm_xxx values work with `yarn start`

Tested locally `docker build + docker run`

Without this branch localhost:3000/env-config.js

`window._env_ = {
  REACT_APP_APPLICATION_NAME: "",
 ...
  REACT_APP_VERSION: "",
}`

When docker image is built + run with this branch:

```
window._env_ = {
  REACT_APP_APPLICATION_NAME: "open-city-profile-ui",
....
  REACT_APP_VERSION: "1.1.11",
}
```


Values were also empty on dev server. Except now that this branch has been deployed:
[https://profiili.dev.hel.ninja/env-config.js](https://profiili.dev.hel.ninja/env-config.js)

Consulted also original author of this env-config + env.sh "bundle".

akikoskinen:
Ah, yes. I missed one little detail in the `env.sh` script. For a line `REACT_APP_APPLICATION_NAME=$npm_package_name` in the `.env` file, it first tries to find an envvar `$REACT_APP_APPLICATION_NAME` and not `$npm_package_name`. That's surprising, but it is what it is.

akikoskinen (approved):

PR #236, opened 2022-01-25
HP-1204: Increase yarn timeout in Dockerfile

NikoHelle:
Build fails in Azure stage and IBM suggested this as a cure.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ USER root
 RUN bash /tools/apt-install.sh build-essential
 
 USER appuser
+RUN yarn config set network-timeout 300000
 RUN yarn && yarn cache clean --force
 
 USER root


akikoskinen (changes requested):
Instead of referring to internal discussions in the commit message, e.g. a link to a [public issue](https://github.com/yarnpkg/yarn/issues/8242) could be added.

NikoHelle:
I could add link to the Slack-discussion if really required
https://helsinkicity.slack.com/archives/C012SR6PHSB/p1643028380074300

I don't think increasing a timeout needs a deeper explanation.

akikoskinen:
No, don't add links or references to any internal sources.

akikoskinen (approved):

PR #237, opened 2022-02-01
Reverted HP-1169

NikoHelle:
Reverted commits starting from HP-1169 merge commit, so other reverted commits were empty and discarded.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "open-city-profile-ui",
-  "version": "1.1.12",
+  "version": "1.1.11",
```
akikoskinen, line 3: Is it okay to go back in version numbers?

NikoHelle: The version number did not change after this, so I think it is logical when reverted to previous version.

NikoHelle: We don't do package releases and version number is only used to for output.

akikoskinen (approved):

PR #238, opened 2022-02-02
HP-1169: translation language to service connections, 3rd version

NikoHelle:
This is almost identical as previous version. It was reverted, but changes in BE did not require many UI changes so same changes again...

Commit before revert: 3a0c04fc335e3e7a07592905fd7916bb070abd74
Last commit in this PR: 325e98011a367d216245c2ed373a78cfac74451d

git diff --name-only 3a0c04fc335e3e7a07592905fd7916bb070abd74..325e98011a367d216245c2ed373a78cfac74451d 

Changed:
generatedTypes.ts 
typings.ts 
ServiceConnectionsQuery.graphql

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "open-city-profile-ui",
-  "version": "1.1.11",
+  "version": "1.1.12",
   "license": "MIT",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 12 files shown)


akikoskinen (changes requested):
Code looks nice. Go through the commit messages and check that the information there is current. Now they seem to also contain snippets from some old solutions that aren't in these commits any more.

NikoHelle:
OK, removed line "DeleteProfile refetched ServiceConnections even if those had never been loaded. Added check for that."

akikoskinen (approved):

PR #243, opened 2022-03-02
HP-1048: country dial code to phone numbers

NikoHelle:
Phone numbers are split into dial code and number. Number part can only have numbers and dial code is "+" with one or more numbers.

It was discussed with the team that number part can  be min 5 chars long. It was also decided that countries with same dial code are combined in the options like: "Canada & USA (+1)"


akikoskinen:
The term "dial code" is used in this PR. It seems to refer to the country code (in telephony terms). But that term could be incorrect. The used terms I can find from the internets are "country calling code", "country dial-in code" or just "country code" (which as such can be mixed with the ISO 3166 country codes, so it's probably not the best option either). The term "dial code" (or "dialling code") seems to be used for the are codes within a country, e.g. https://en.wikipedia.org/wiki/List_of_dialling_codes_in_the_United_Kingdom. I suggest to use a more correct term.

NikoHelle:
Yeah, "countryCode" was already taken, so "dialCode" was better. And shorter than "countryCallingCode", "countryDialIn Code". "DialCode" is not actually correct, but I thought the best option and clear enough.

I'll change it then to "countryCallingCode". Is that ok?

NikoHelle:
DialCode is replaced now with "countryCallingCode". I'll push when tests go through.

NikoHelle:
Renamed props in all commits and push is complete.

akikoskinen (changes requested):
A few usability issues I encountered.

1) In phone edit mode, open the country code dropdown and select one of the options. Open the dropdown again. Now there's only the previously selected option available. The user needs to delete characters from the country code input in order to see more options again.

2) In phone edit mode, type into the country code input for example `+358`. Unfocus the country code input, for example by selecting the number input. The country code gets cleared although the user inserted a valid country code there.

3) In phone edit mode, ensure some option is selected as the country code. Change UI language. The selected option's language doesn't change.

NikoHelle:
I find the combobox with search field quite annoying too. With about 200 country options, the search must be enabled.

To the first and second issue: that's how the HDS component works: 
https://hds.hel.fi/storybook/react/?path=/story/components-phoneinput--with-country-code
https://hds.hel.fi/storybook/react/?path=/story/components-dropdowns-combobox--combobox-example

We can start to override the functionalities, but they have done a lot of testing and accessibility iteration, so it should work correctly.

The idea is probably that typing is just for searching and user MUST select an item.
  
 Translation issue can be fixed.

akikoskinen:
> I find the combobox with search field quite annoying too. With about 200 country options, the search must be enabled.
> 
> To the first and second issue: that's how the HDS component works: https://hds.hel.fi/storybook/react/?path=/story/components-phoneinput--with-country-code https://hds.hel.fi/storybook/react/?path=/story/components-dropdowns-combobox--combobox-example
> 
> We can start to override the functionalities, but they have done a lot of testing and accessibility iteration, so it should work correctly.
> 
> The idea is probably that typing is just for searching and user MUST select an item.

Yeah, no point in trying to circumvent the HDS components' problems in one project. Perhaps inform the HDS project that there could be some room for improvement.

akikoskinen (approved):

PR #246, opened 2022-03-17
HP-908 remove useless profile exists check

NikoHelle:
The existence of a profile can be checked just by loading the user's profile. So the profile exists check is skipped now and removed completely.

Checking if user...
- is authenticated
- is loading a profile 
- has a profile 
- profile is loaded, but does not exist
- profile load failed
- authentication failed

...is a bit complicated. And due the nature of a SPA, user may navigate back and forth after any loading has started. Which makes checks even more complicated.

So added a hook which tracks the different scenarios reliably and with comprehensive tests.

### DIFF
--- a/src/profile/components/profile/Profile.tsx
+++ b/src/profile/components/profile/Profile.tsx
@@ -1,8 +1,6 @@
 import React from 'react';
-import { ApolloError } from '@apollo/client';
 import { useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router';
-import * as Sentry from '@sentry/browser';
 import { Button, Notification } from 'hds-react';
 
 import PageLayout from '../../../common/pageLayout/PageLayout';
@@ -12,30 +10,18 @@ import Loading from '../../../common/loading/Loading';
 import styles from './Profile.module.css';
 import authService from '../../../auth/authService';
 import responsive from '../../../common/cssHelpers/responsive.module.css';
-import { useProfileErrorListener } from '../../context/ProfileContext';
-import parseGraphQLError from '../../helpers/parseGraphQLError';
-import useToast from '../../../toast/useToast';
 import { useProfileLoadTracker } from '../../hooks/useProfileLoadTracker';
 import { WithAuthCheckChildProps } from '../withAuthCheck/WithAuthCheck';
 
 function Profile(props: WithAuthCheckChildProps): React.ReactElement {
   const { t } = useTranslation();
   const location = useLocation();
-  const { createToast } = useToast();
   const {
     hasExistingProfile,
     isProfileLoadComplete,
     didProfileLoadFail,
     reloadProfile,
   } = useProfileLoadTracker();
-
-  useProfileErrorListener((apolloError: ApolloError | Error) => {
-    if (parseGraphQLError(apolloError).isAllowedError) {
-      return;
-    }
-    Sentry.captureException(apolloError);
-    createToast({ type: 'error' });
-  });
 
   const getPageTitle = () => {
     const pathname = location.pathname.substr(1);
--- a/src/profile/components/profile/Profile.tsx
+++ b/src/profile/components/profile/Profile.tsx
@@ -1,9 +1,7 @@
-import React, { useContext, useEffect, useState } from 'react';
-import { useLazyQuery, ApolloError } from '@apollo/client';
+import React from 'react';
+import { ApolloError } from '@apollo/client';
 import { useTranslation } from 'react-i18next';
-import { useHistory, useLocation } from 'react-router';
-import { loader } from 'graphql.macro';
-import { User } from 'oidc-client';
+import { useLocation } from 'react-router';
 import * as Sentry from '@sentry/browser';
 import { Button, Notification } from 'hds-react';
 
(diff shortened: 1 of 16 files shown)


akikoskinen (commented):
It confuses me why authentication needs to be mixed with profile fetching. They are unrelated, although dependent concepts. Authentication is a prerequisite for profile loading, but that doesn't  mean that they have to be tightly coupled with each other.

To me it seems that the logic could be something like: 
1. Handle authentication. If not authenticated, ask to authenticate. Handle errors in authentication. Eventually this leads to an authenticated user (or the user goes away, if they can't or don't want to authenticate).
2. After authenticated, start dealing with profile. See if there's a profile present. If not, create one. If yes, show it. Handle errors in fetching.

The execution would go to step 2 only after step 1 has completed successfully. The logic of step 1 would be implemented in it's own module and the logic of 2 would be in another module. ("Module" being a hook or whatever is the correct tool for this).

There shouldn't be any modules (hooks, files, etc.) that has both authentication and profile related stuff, except for one place where there would be something like:

```jsx
if (isAuthenticated()) {
    return <Profile/>;
}
```

NikoHelle:
You're right. Would simplify the process if those where separate. I tried to make a small change, but it got bloated and still tried kept the same components.

src/profile/hooks/useProfileLoadTracker.ts
```diff
@@ -0,0 +1,139 @@
...
+  return STATUS_LOADING;
+}
+
+export function didLoadFail(currentStatus?: Status): boolean {
+  return currentStatus === STATUS_FAILED;
+}
+
+export function getLoadFailedStatus(): Status {
+  return STATUS_FAILED;
+}
```
akikoskinen: It doesn't seem too useful to have this kind of small helpers that just basically give another name for a simple statement. Even if the new name is useful as an internal detail, I don't see why these should be exported.

src/profile/hooks/useProfileLoadTracker.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }, [newStatus, updateStatus, profileContext]);
+
+  return {
+    isProfileLoadComplete: () => isLoadComplete(valueRef.current),
+    didProfileLoadFail: () => didLoadFail(valueRef.current),
+    hasExistingProfile: () => {
+      if (!isLoadComplete(valueRef.current)) {
+        throw new Error(
+          'hasExistingProfile cannot be used before profile load is complete. Check it with isLoadComplete()'
```
akikoskinen:
```suggestion
          'hasExistingProfile cannot be used before profile load is complete. Check it with isProfileLoadComplete()'
```

src/profile/hooks/useProfileLoadTracker.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (canStartLoading(currentStatus) && !shouldLoad(currentStatus)) {
+    return getShouldLoadStatus();
+  }
+
+  return undefined;
+}
+
+export function useProfileLoadTracker(): useProfileLoaderHookReturnType {
+  const profileContext = useContext(ProfileContext);
+  const valueRef = useRef<Status>(STATUS_WAITING);
```
akikoskinen: The value's name should be something more explanatory. Perhaps something like `currentStatus`.

src/profile/components/profile/Profile.tsx
```diff
@@ -93,60 +39,64 @@ function Profile(): React.ReactElement {
...
+          tunnistamoUser={props.user}
+          onProfileCreated={() => reloadProfile()}
+        />
+        ;
       </PageLayout>
     );
   }
 
   return (
-    <PageLayout title={getPageTitle()} disableFocusing>
```
akikoskinen, line 132: Should the `disableFocusing` property still be present somewhere in this `Profile` component?

akikoskinen (changes requested):
Some smallish notes. Generally looks pretty good.

src/profile/hooks/useProfileLoadTracker.ts
NikoHelle: Fixed!

src/profile/hooks/useProfileLoadTracker.ts
NikoHelle: Changed to "statusRef", because there is also "updateStatus", not "updateCurrentStatus" and currentStatusRef.current sounds weird :)

src/profile/components/profile/Profile.tsx
NikoHelle: True. Added to loader in WithAuthCheck too

src/profile/hooks/useProfileLoadTracker.ts
NikoHelle: I agree. But they have to be exported for tests. Other option is to export all STATUS_xxx constants, but I think this is cleaner.

akikoskinen: I would prefer exporting the `STATUS_*` constants. They kind of form an enum and are thus grouped to a whole, whereas a bunch of functions are just separate functions without that much connection to each other. But JS/TS is weird to me anyway, so perhaps instead of a neat enum it's customary to create a bunch of oneliner functions :shrug:

NikoHelle: Changed to use STATUS_*

src/profile/hooks/useProfileLoadTracker.ts
```diff
@@ -0,0 +1,119 @@
...
+  if (profileContext.data && !isLoadedSuccessfully(currentStatus)) {
+    return STATUS_SUCCESS;
+  }
+
+  if (canStartLoading(currentStatus) && !shouldLoad(currentStatus)) {
+    return STATUS_SHOULD_LOAD;
+  }
+
+  return undefined;
+}
```
akikoskinen: All these functions are internal implementation details and shouldn't be exported. These are now only used in tests, not in production code. It's better to test the _behaviour_, not _implementation_. So test through the public API.

src/profile/hooks/useProfileLoadTracker.ts
```diff
@@ -0,0 +1,119 @@
...
+  | typeof STATUS_SUCCESS
+  | typeof STATUS_SHOULD_LOAD;
+
+export type useProfileLoaderHookReturnType = {
+  isProfileLoadComplete: () => boolean;
+  hasExistingProfile: () => boolean;
+  didProfileLoadFail: () => boolean;
+  reloadProfile: () => Promise<QueryResult>;
+  getStatus: () => Status;
+  getProfile: () => ReturnType<ProfileContextData['getProfile']>;
```
akikoskinen: The `getStatus` and `getProfile` functions don't seem to be used in any production code, so can they be removed? If `getStatus` is not needed, perhaps the `STATUS_*` consts and `Status` type don't have to be exported either.

akikoskinen (changes requested):
`useProfileLoadTracker` still seems to reveal too much of its private parts.

src/profile/hooks/useProfileLoadTracker.ts
NikoHelle: Removed implementation tests and unused exports. Added one behavioural test.

In React Apps behavioural tests are quite hard to implement so they are clear and cover all cases. Doable of course, but faking API requests and responses, hooks, dom, user behaviour etc makes tests confusing to read and it takes time to find out if everything is also faked correctly. And all the fake stuff must also be maintained.

Easier to expose and test "safe" parts like resolveChangedStatus() which just returns new status and does not change internal status.

akikoskinen (approved):
Nice

PR #248, opened 2022-03-21
HP-1335: indicate required fields

NikoHelle:
Added (*) to form fields which are required in address and basic data forms.

Didn't add them to phone and email fields because the is no need for extra descriptions.

Screen readers already have the information so the note has "aria-hidden"

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -232,7 +232,8 @@
     "required": "This field is required.",
     "errorTitleForScreenReaders": "The form has the following errors:",
     "countryCallingCodeRequiredIfNumber": "Country code is required with phone number",
-    "numbersOnly": "Only numbers allowed"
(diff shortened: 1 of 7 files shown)


src/i18n/en.json
```diff
@@ -232,7 +232,8 @@
     "required": "This field is required.",
     "errorTitleForScreenReaders": "The form has the following errors:",
     "countryCallingCodeRequiredIfNumber": "Country code is required with phone number",
-    "numbersOnly": "Only numbers allowed"
+    "numbersOnly": "Only numbers allowed",
+    "requiredFieldIndication": "Fields marked with an * are compulsary."
```
akikoskinen: s/compulsary/compulsory/.

akikoskinen (changes requested):

akikoskinen (approved):

PR #249, opened 2022-03-24
HP-1337: set page focus

NikoHelle:
Some screen readers place focus to a random element when SPA renders new content. This was reported in accessibility audit.

Page focus is now controlled and set when a new page is rendered. The focused element is usually h1. User's name used to be h1 on Profile page, so had to change that, because it is useless heading for screen readers. That change forced to change other headings too to keep them in order.

Focus is **not** set on first page load, because focus would jump over heading etc which are important on first load. There is "skip to content" -link anyway. 

Sometimes the app redirects the user, so those redirects must also be taken in to account, when determining is current page the first one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "open-city-profile-ui",
-  "version": "1.1.14",
+  "version": "1.1.15",
   "license": "MIT",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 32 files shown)


akikoskinen (changes requested):
Some issues with the commit order.

Commit 7d423a7d22 uses `FocusableH1` (in tests) but that component is introduced only in a later commit b1f5f6013ed.

Commit 82872448c64 passes `disableFocusing` to `PageLayout` but `PageLayout` learns to accept that property only in a later commit 36096b11a0.

As a result quite a few commits in this PR fail to compile (either the app or tests). I suggest massaging the commits so that every commit builds.

akikoskinen (approved):

PR #254, opened 2022-04-21
HP-1382 single address only

NikoHelle:
Users should only have max one address. Old users may have more, but new ones cannot be added. All old ones can be removed.

This is a draft for Sonar checkups.


NikoHelle:
Closed until decided is this enough.

PR #256, opened 2022-04-27
HP-1320: update packages

NikoHelle:
Main purpose was to update react-scripts, but it required a lot of other updates.

React-scripts required newer node, so updated from 12 to 14.

HDS was also updated. It is not fully compatible with react-scripts v5, but fixed by manually adding missing packages.

New react-scripts broke linting and while fixing it added also linters used in HDS so we have same rules.

New rules raised some linting error and fixed those.

Updated also some packages to latest versions. Apollo/client's latest version does not work with our codebase. Not even minor version update, so locked it to current version. 

Draft until build is successful on the server side.


akikoskinen (approved):
:+1:

PR #259, opened 2022-05-03
HP-1382: user can have only one address

NikoHelle:
Essential changes:
- users may add only one address 
- addresses cannot be set as primary
- all addresses can be removed

Users may already have saved multiple addresses, so all are shown and can be edited and removed.

Local version can be tested by adding multiple addresses in profile DEV.


akikoskinen (changes requested):
Looking quite nice. The only issue I can think of is the commits related to translations.

Commit 340fe69e29 starts using new translations that are only added in a later commit, 0f011b673. Also the commit 0f011b673 removes some translations that are used until a yet later commit 613503787. As a result quite a many of the commits are not fully functional because not all used translations are available.

NikoHelle:
I thought it was clearer to split changes into multiple commits rather than put all changes with dependencies to translations into one commit.

I'll split the translations into commits that use the changed parts.

NikoHelle:
Translation changes are now split in to the commits.

akikoskinen (approved):

PR #261, opened 2022-05-16
DEVOPS-402 CPU Throttling high alerts

terovirtanen:
Increase cpu limit


nikomakela (approved):
If it's throttling, the limits might be good to be increased, but please note that the KUVA instance might not even be in use. At least the Berth-reservations currently uses the one in Kanslia's Azure.

NikoHelle:
Only master branch is deployed to non-azure environments. So this should not have any affect at the moment.

PR #263, opened 2022-05-18
HP-1425: fix type overwrite

NikoHelle:
The emailType, addressType and phoneType in save data was always reset to the default value "OTHER" when saving. The value is now stored in editData ja inserted to the saved data.

Fixed also a css problem when error message was shown with phone number. That resulted some code cleaning.


akikoskinen (commented):
It surprises me a little that this change needs so much _new_ code in `editData.ts`. I would have expected that it's enough to _remove_ code that always forces the `type` to `OTHER` (there must be that code somewhere). I'm thinking that the logic for `type` would be exactly the same that for any other piece of editable data: the data goes to the UI, the user can edit it, the modified data comes back to `editData.ts`, data gets sent to backend.

In the case of `type` attribute the user of course can not really edit it because it's not visible in the UI. But from the point of `editData.ts`'s view, that shouldn't make a difference.

NikoHelle:
As stated in the ticket, the "OTHER" must be sent when creating new items. Just removing all type-properties from sent nodes would have been quite easy, but there must be a default for each phone/email/address type.
If new items would not require the type, then it could just be deleted from all queries in Profile UI as it does not use it.

Defaults and mutation variables for GraphQL queries are created by` updateMutationVariables.ts `

I don't think that is the correct place to filter out data not shown in UI. `EditData` knows what properties are editable in the UI and the `updateMutationVariables` should just sent what is passed to it. That's why `editData` now stores also the unedited data and `updateMutationVariables` is unchanged. 

When updating data, the `updateMutationVariables` could remove all properties that are null/undefined. That could fix this too.

akikoskinen:
> Defaults and mutation variables for GraphQL queries are created by`updateMutationVariables.ts`
> 
> I don't think that is the correct place to filter out data not shown in UI. `EditData` knows what properties are editable in the UI and the `updateMutationVariables` should just sent what is passed to it.

I was thinking that `EditData` shouldn't care either what the UI shows (or what is editable there) and what it doesn't. `EditData` could be "prepared" for manipulating the `type` attribute as well. When somebody figures in the future, that actually they user _should_ be able to control the `type` too, then it would be only a change in the UI. The data layer would already be ready to handle such changes.

NikoHelle:
EditData was created when there was a need for adding multiple addresses, emails and phones. All at the same time. And addresses have multiple fields. Plus all of items could be set as primary. 

In React world there is no way the component (UI) itself could handle simultaneous editing, adding and removal of multiple items - as is still required.

EditData cannot be just an arbitrary object where UI puts stuff.
There are many scenarios, where values and props must be known
- a changed value(s) must be detected (yes, ui could add a flag for itself)
- data must be resettable if user cancels (yes, ui could handle it, but why, if storage (editData) exists)
- an existing new item must be known, because 2 new items cannot be added simultaneously (yes, obsolete at the moment)
- primary items must be set and unset in sync (yes, obsolete at the moment)
- etc.

If some day types must be selected, then there might a need for preventing multiple items of same type, so there must be a centralized place to store items. 

Within current specs, the editData could be rewritten as a dumber storage. So many features have been dropped. Specs will probably change again 😄

PS. This PR enables EditData to handle types.

akikoskinen:
Perhaps I'm starting to get this. I was thinking that `editData` would directly represent the data model, offering all editing possibilities that the data model enables. But now I see that's not the case. It already seems to have opinions about what can be edited, limiting some editing possibilities to the needs of the current UI. So `editData` seems to be somewhere in between what the data model enables and what UI needs. Maybe I'm a bit lost _where exactly_ in between it resides.

This `SaveDataMutator` looks weird to me. Why is it needed to modify the data coming from the editor (the UI) before really saving it? It wasn't needed before so why is it needed now? Is it because the `type` isn't part of `EditDataValue`? Because I'm thinking that perhaps it should be. These `type` attributes for emails, phones and addresses are similar to the other editable attributes: just like e.g. `postalCode` in an address. So shouldn't the various `type`s be included into `EditDataValue` (via `AddressValue`, `PhoneValue` and `EmailValue`) and not into `EditData` as a new readonly attribute?

NikoHelle:
The editing process before this PR:
- get data from BE
- parse and store the data UI needs to editData: `id`, `primary`, `value` (= input fields: names / email / phones / addresses)
- `emailType`, `phoneType`, `addressType` is not stored as UI does not use it.
- when user changes values in inputs, update `editData.value`  (primary cannot be changed anymore)
- when user hits save, create save data by calling `createFormValues()`
- the function creates graphQL nodes from editData: `id`, `primary`, and all fields from `editData.value`
- emailType, phoneType, addressType is not stored, so created nodes do not have it
- `updateMutation.ts` creates a GraphQL mutation object. Types are not found -> sets them as default . This is legacy code and haven't been changed.

In this PR
- editData stores the type which is picked from `emailType`, `phoneType`, `addressType`
- when saving, `saveDataMutator` sets the stored type to the save data
- `updateMutation.ts` finds the values and does not set defaults

SaveDataMutator is a middleware which merges the data shown in the UI and data not shown, so all the data is still present in the save object.

All fields are not in the `editData.value`, because all are not editable in input fields. React form-component (Formik) uses only the `editData.value` and its properties and typings. If `primary` and `type` would be in `editData.value`, it would force to handle them also in the Formik-component. At least pass them through. 

Nothing would be gained by that, because graphQL nodes are still created from editData.

Answers:
_"editData seems to be somewhere in between what the data model enables and what UI needs. Maybe I'm a bit lost where exactly in between it resides."_

It picks the editable props from graphQL nodes. Well, `primary` is not editable anymore. "Why" is explained above and previous answer regarding why editData exists.

_"It <SaveDataMutator> wasn't needed before so why is it needed now?"_

Because `updateMutation` sets default values to the type-fields when value is missing. "Just remove that!" It has to be there when creating new nodes.

 _"So shouldn't the various types be included into EditDataValue (via AddressValue, PhoneValue and EmailValue) and not into EditData as a new readonly attribute?"_

No, the `value` is used in forms and should not have other data than used in input fields. 
Why does it matter if `type` was there, unlike `primary`?

akikoskinen:
> All fields are not in the `editData.value`, because all are not editable in input fields. React form-component (Formik) uses only the `editData.value` and its properties and typings. If `primary` and `type` would be in `editData.value`, it would force to handle them also in the Formik-component. At least pass them through.

Aha, this part I was missing. Now I understand why `editData` needs to provide data exactly how it's shown in the UI. Perhaps this then needs to be done this way, if "pass them through" needs some extra handling and isn't thus an option.

akikoskinen (approved):

PR #264, opened 2022-06-01
HP-1377: handle multiple graphql errors with download and delete profile.

NikoHelle:
The original error seems to resolve after updating Apollo/client. Errors are shown multiple times now. Tested with Chrome and network throttling.

Added missing error handling to service connection fetching. 

Also changed displaying errors in download -component to match other Profile components: notification instead of toast.

While testing found a new bug: while saving emails, the save success message was not shown anymore. Fixed that.

### DIFF
--- a/src/profile/components/emailEditor/EmailEditor.tsx
+++ b/src/profile/components/emailEditor/EmailEditor.tsx
@@ -212,6 +212,9 @@ function EmailEditor(): React.ReactElement | null {
           {t('profileForm.addEmail')}
         </Button>
       )}
+      {!willSendEmailVerificationCode && (
+        <EditingNotifications content={content} dataType={dataType} />
+      )}
     </ProfileSection>
   );
 }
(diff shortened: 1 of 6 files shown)


src/profile/components/emailEditor/EmailEditor.tsx
```diff
@@ -212,6 +212,9 @@ function EmailEditor(): React.ReactElement | null {
           {t('profileForm.addEmail')}
         </Button>
       )}
+      {!willSendEmailVerificationCode && (
+        <EditingNotifications content={content} dataType={dataType} />
+      )}
```
akikoskinen, line 217: This is good for now. Just a note for future: this will need changes once the username/password authentication option will be taken into use. The email verification will happen for any user who has logged in via Tunnistus, no matter if it was using suomi.fi or some other means.

akikoskinen (approved):

src/profile/components/emailEditor/EmailEditor.tsx
NikoHelle: Ok, that is a good change.

PR #271, opened 2022-08-15
HP-1468: removed account links

NikoHelle:
Link to the used authorisation provider (Google, Suomi.fi) was removed from the UI.


akikoskinen (approved):
:+1:

PR #272, opened 2022-08-31
HP-1473: new texts and styles

NikoHelle:
Updates to texts and text layouts.

This PR does not include major UI changes like "unboxing" Download and Delete Profile components or padding and colour changes. Those are in the Figma-layout linked in the Jira-ticket, but will be done in another PR.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -184,7 +184,6 @@
     "permanentAddress": "Permanent address",
     "permanentForeignAddress": "Permanent foreign address",
     "verifiedDataInformation": "You can only see the official information when you have logged in using strong authentication via the Suomi.fi service. The information has been retrieved from The National Population Information System and cannot be edited via your Helsinki profile. You can check the data stored about you in the Population Information System in the <suomiFiLink>suomi.fi -web service</suomiFiLink>.",
-    "verifiedDataInformationForScreenReaders": "EN: The data below have been confirmed from the national Population Information System and therefore cannot be edited in your Helsinki profile. If you want, you can view your personal data stored in the Population Information System in the suomi.fi web service: <suomiFiLink>go to the suomi.fi service </suomiFiLink>",
     "verifiedDataInformationLink": "https://testipalvelu.apro.tunnistus.fi",
     "verifiedData": "Verified data",
     "verifiedBasicData": "Official information",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -35,7 +35,7 @@
   "deleteProfile": {
     "accept": "I understand what deleting my data means and wish to proceed.",
     "delete": "Delete my information",
-    "explanation": "If you want, you can delete all the data we hold on you. Once deleted, the data cannot be restored. If you have linked your profile to a City of Helsinki service, you will no longer be able to use that service until you create a new profile. After the deletion, you will be automatically logged out of this service. If you have used a third-party service provider account (such as Google or Facebook) to log in, that account will not be deleted.",
+    "explanation": "If you want to, you can delete your Helsinki profile and all the data we have stored. Once deleted, the data cannot be restored. A new Helsinki profile must be created to use the e-services associated with your Helsinki profile.<br/><br/>After the deletion, you will be automatically logged out of this service.<br/><br/>If you have logged in using a third-party service provider account (e.g. Google or Facebook), that account will not be deleted.",
     "title": "Do you want to delete your information?",
     "loadingServices": "Searching for services connected to the profile"
   },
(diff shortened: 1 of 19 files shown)


src/i18n/en.json
```diff
@@ -167,7 +172,7 @@
     "ariaSelectedOption": "{{value}} is selected",
     "ariaNoSelectedItemForLabel": "{{label}} is not selected",
     "deleteProfile": "Delete your profile",
-    "description": "The Helsinki profile enables you to use your data in all City of Helsinki online services that require authentication. One set of data, easily accessible on various services.",
+    "description": "The data stored in the Helsinki profile is used in the City of Helsinki’s e-services. You can find more detailed information <linkToServices>{{linkToServicesText}} on the tab</linkToServices>.",
```
akikoskinen: The end becomes "You can find more detailed information Your services on the tab". Should probably be changed to something better.

src/i18n/sv.json
```diff
@@ -167,22 +172,21 @@
     "ariaSelectedOption": "{{value}} har valts",
     "ariaNoSelectedItemForLabel": "{{label}} har inte valts",
     "deleteProfile": "Radera profil",
-    "description": "Helsingfors-profilen gör det möjligt att använda dina uppgifter i alla Helsingfors webbtjänster som kräver identifiering. Samma information, enkel att använda i många olika tjänster.",
+    "description": "Uppgifterna som sparats i Helsingforsprofilen används i Helsingfors stads e-tjänster. Närmare information hittar du <linkToServices>{{linkToServicesText}} på fliken</linkToServices>.",
```
akikoskinen: Samma problem här. Texten blir "Närmare information hittar du Dina tjänster på fliken".

src/i18n/fi.json
```diff
@@ -33,9 +33,9 @@
...
-    "accept": "Ymmärrän mitä tietojeni poistaminen tarkoittaa ja haluan jatkaa.",
+    "accept": "Ymmärrän, mitä tietojeni poistaminen tarkoittaa ja haluan jatkaa.",
     "delete": "Poista omat tiedot",
-    "explanation": "Jos haluat, voit poistaa kaikki sinusta tallentamamme tiedot. Tietoja ei voi palauttaa sen jälkeen, kun ne on poistettu. Jos olet yhdistänyt profiilisi johonkin Helsingin kaupungin palveluun, et enää pysty käyttämään kyseistä palvelua ennen kuin luot uuden profiilin. Poistamisen jälkeen sinut kirjataan automaattisesti ulos tästä palvelusta. Jos olet käyttänyt kirjautumisessa ulkopuolisen palveluntarjoajan tiliä (esim. Google tai Facebook), kyseistä tiliä ei poisteta. ",
+    "explanation": "Halutessasi voit poistaa Helsinki-profiilisi ja kaikki sinusta tallentamamme tiedot. Tietoja ei voi palauttaa sen jälkeen, kun ne on poistettu. Helsinki-profiiliisi yhdistettyjä asiointipalveluja käyttääksesi on luotava uusi Helsinki-profiili.<br/><br/>Poistamisen jälkeen sinut kirjataan automaattisesti ulos tästä palvelusta.<br/><br/>Jos olet käyttänyt kirjautumisessa ulkopuolisen palveluntarjoajan tiliä (esim. Google tai Facebook), kyseistä tiliä ei poisteta.",
```
akikoskinen, line 38: Would it be more correct to enclose paragraphs within `p` elements?

src/i18n/fi.json
```diff
@@ -159,6 +160,10 @@
   },
   "profileInformation": {
     "address": "Osoite",
+    "addressDescription": "Nämä tiedot olet itse lisännyt Helsinki-profiili -palveluun tai muuhun Helsingin kaupungin asiointipalveluun. Tätä osoitetietoa käytämme silloin, kun olet tunnistautunut kevyesti, eli jotain muuta kuin Suomi.fi -tunnistautumista hyödyntäen. Voit halutessasi poistaa tiedot Helsinki-profiilista, jolloin ne poistuvat myös muista asiointipalveluista. Älä poista tietoja, jos sinulla on asiointi kesken.",
+    "addressDescriptionNoWeakAddress": "Et ole lisännyt toista osoitetta. Toista osoitetietoa käytämme silloin, kun olet tunnistautunut kevyesti, eli jotain muuta kuin Suomi.fi -tunnistautumista hyödyntäen.",
+    "addressDescriptionNoAddress": "Et ole lisännyt osoitetta. Tätä osoitetietoa käytämme silloin, kun olet tunnistautunut kevyesti, eli jotain muuta kuin Suomi.fi -tunnistautumista hyödyntäen.",
```
akikoskinen, line 165: IMO these explanations are somewhat misleading. What address a service uses is not dependent on the user's (current) authentication method. Basically services can use the addresses in whatever way they want, only depending on what addresses a profile contains.

akikoskinen (changes requested):
There are problems with the commit contents and/or order. Translation texts are removed before their usage is removed. Styles are used before they are introduced.

src/i18n/en.json
NikoHelle: Changed!

src/i18n/fi.json
NikoHelle: True, but it was discussed that there is no need for deeper explanation.

src/i18n/fi.json
NikoHelle: The text is inserted in to a &lt;p&gt; element in the code. &lt;br&gt; is better in this case between text blocks.

akikoskinen (approved):

PR #274, opened 2022-09-21
HP-1526: New layout

NikoHelle:
New layout based on https://www.figma.com/proto/WaYKxyMHA5Gze1lmJICLbI/%F0%9F%91%A4-Helsinki-profiili?page-id=0%3A1&node-id=2%3A2&viewport=899%2C639%2C1&scaling=min-zoom

Biggest changes are:
- white boxes are gone
- download and delete profile boxes are not accordions anymore (can't open/close)
- bottom page under download and delete profile components is gray


akikoskinen:
The service connections are fetched from BE many times unnecessarily. Click "delete profile" (but don't actually delete the profile). Go to "Your services" tab, return to information tab. Go to "delete profile" again (without deleting it this time either). The service connections are fetched many times during these steps while one time would be enough.

NikoHelle:
I had to add `fetchPolicy: 'no-cache'` to line 77 in `DeleteProfile.tsx`. This causes reload every time user navigates to the Profile -page and clicks the delete button. (Which is not very often).

Without "no-cache", the Apollo-client does not trigger `onComplete`-callback, if data is already in cache.

Same has been done before in this commit:
https://github.com/City-of-Helsinki/open-city-profile-ui/commit/4f1d9de6dba7aaef51890e2d58b6459da36b1cdc

Previously caching did not matter, because service connections were loaded before user can press the delete button and confirm the deletion from the confirmation modal.

Now service connections are loaded after delete is pressed and confirmation modal is showed after that. When cached, the modal would never been show as `onComplete`-callback would not been called.

Very annoying. At least service connections are not fetched every time delete is pressed. Only if user navigates away from that page and tries to delete again.

akikoskinen:
Click "delete profile", but cancel the modal. Change language. Profile deletion modal pops up again.

NikoHelle:
The "onComplete"-callback is called again when component is re-rendered after language change. Happens when serviceConnections have already been fetched.

Added a check for that and it is not happening anymore.

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -67,13 +67,21 @@ function DeleteProfile(): React.ReactElement {
...
     ServiceConnectionsQueryVariables
   >(SERVICE_CONNECTIONS, {
     variables: createServiceConnectionsQueryVariables(i18n.language),
+    fetchPolicy: 'no-cache',
     onCompleted: () => {
+      if (dataLoadState === loadedLoadState) {
+        return;
+      }
       setDataLoadState(loadedLoadState);
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
akikoskinen: Is the `@typescript-eslint/no-use-before-define` rule something that should be followed (i.e. reorganize definitions above usages), or can the rule just be disabled in eslint configuration?

src/auth/components/login/Login.module.css
```diff
@@ -4,6 +4,9 @@
   align-items: center;
   flex-direction: row;
   flex: 1;
+  background-color: var(--color-black-ultra-light);
+  /* footer koros is 14px high and "bleeds" bg color. This shifts the koros up and hides the bleed */
+  margin-bottom: -14px;
```
akikoskinen: Login page has this negative margin to put the background color "between the waves" of the footer koros. But the profile information page doesn't have this. There the "between the waves" is white, while the bottommost parts (delete profile) of the main content background color is light gray.

src/profile/components/deleteProfile/DeleteProfile.tsx
```diff
@@ -149,48 +154,37 @@ function DeleteProfile(): React.ReactElement {
...
-              className={styles.button}
-              id={removeButtonId}
-            >
-              {t('deleteProfile.delete')}
-            </Button>
-          </div>
-        )}
-      </ExpandingPanel>
-
+    <ProfileSection hasVerifiedUserData data-test-id={'delete-profile'}>
```
akikoskinen: I suppose the `hasVerifiedUserData` value has lost its original meaning. Now it seems to be used to mean "use `background: none` CSS style". The value and perhaps also the `.verified-data` class should probably be renamed.

or

If `hasVerifiedUserData` is `false`, then `ProfileSection` gets a white background. This seems to be used in places where the underlying background color is also white, so it doesn't add anything visible. Can the white background and `hasVerifiedUserData` flag just be removed from `ProfileSection`?

akikoskinen (changes requested):

src/profile/components/deleteProfile/DeleteProfile.tsx
NikoHelle: True, good catch. Removed the css and usages.

src/auth/components/login/Login.module.css
NikoHelle: Fixed! Moved the margin to base element of all content areas

src/profile/components/deleteProfile/DeleteProfile.tsx
NikoHelle: Sometimes functions cannot be ordered by usage, because they cross-use each other and one usage is always before the definition. 

In this case re-ordering can be done

src/profile/components/deleteProfile/DeleteProfile.tsx
akikoskinen: Shouldn't the white background now be removed from the `.profile-section` style, making `ProfileSection`s transparent?

NikoHelle: Yes, and also useless border-radius is now also removed. 

Removed also TopSectionWithKoros component because it is not used anywhere.

akikoskinen (approved):

PR #276, opened 2022-11-08
HP-1657: delete single service connection

NikoHelle:
Added service connection deletion to UI.

Also combined fetching service connections and the authorization code as the same process is now used in three places: download and delete profile and delete service connection.

This branch is currently deployed in dev.

Note about errors:
- if a network or a graphQL error is returned, then a generic error text "Error occured. Try again later" is shown
- if the deletion request succeeds, but returned data has success: false, another error is displayed saying "the data could not be deleted, contact the service to delete them".

To test this, you can add and remove service connections from a user in Profile BE. Only "Esimerkkipalvelu & UI" and "Test service GDPR deletion" can be deleted. Other ones do not have GDPR apis in dev.

"Esimerkkipalvelu" will return error, if pet name is set to "nodelete"

"Test service GDPR deletion" is using free API testing tool "beeceptor.com" as an endpoint. It has daily query limit of 50, so it might return status 429 anytime.

### DIFF
--- a/src/common/expandingPanel/ExpandingPanel.tsx
+++ b/src/common/expandingPanel/ExpandingPanel.tsx
@@ -1,4 +1,9 @@
-import React, { PropsWithChildren, useEffect, useRef } from 'react';
+import React, {
+  PropsWithChildren,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+} from 'react';
 import {
   Button,
   Card,
@@ -41,9 +46,16 @@ function ExpandingPanel({
     container.current = ref;
   };
 
-  const { isOpen, buttonProps, contentProps } = useAccordion({
+  const { isOpen, buttonProps, contentProps, openAccordion } = useAccordion({
     initiallyOpen,
   });
+
+  useLayoutEffect(() => {
+    if (initiallyOpen && isOpen !== initiallyOpen) {
+      openAccordion();
+    }
+  }, [initiallyOpen, openAccordion, isOpen]);
+
   useEffect(() => {
     if (onChange) {
       onChange(isOpen);
--- a/src/gdprApi/useServiceConnectionsAuthorizationCode.ts
+++ b/src/gdprApi/useServiceConnectionsAuthorizationCode.ts
@@ -2,7 +2,7 @@ import { useLazyQuery, ApolloError } from '@apollo/client';
 import { loader } from 'graphql.macro';
 import { useRef } from 'react';
 
-import { GdprServiceConnectionsRoot } from '../graphql/typings';
+import { GdprServiceConnectionsRoot, Service } from '../graphql/typings';
 import useAuthorizationCode from './useAuthorizationCode';
 import { getDeleteScopes, getQueryScopes } from './utils';
 
@@ -20,6 +20,7 @@ export type LoadStatus = {
 type Options = {
   requiredGdprScope: 'delete' | 'query';
   deferredAction: string;
+  serviceName?: Service['name'];
   onError?: (e: ApolloError, status: LoadStatus) => void;
   onCompleted?: (authorizationCode: string, status: LoadStatus) => void;
 };
@@ -71,8 +72,8 @@ function useServiceConnectionsAuthorizationCode(
       onCompleted: data => {
         const scopes =
           options.requiredGdprScope === 'query'
-            ? getQueryScopes(data)
-            : getDeleteScopes(data);
+            ? getQueryScopes(data, options.serviceName)
+            : getDeleteScopes(data, options.serviceName);
         getAuthorizationCode(scopes);
       },
       onError: error => {
--- a/src/gdprApi/useServiceConnectionsAuthorizationCode.ts
+++ b/src/gdprApi/useServiceConnectionsAuthorizationCode.ts
@@ -1,0 +1,107 @@
+import { useLazyQuery, ApolloError } from '@apollo/client';
+import { loader } from 'graphql.macro';
+import { useRef } from 'react';
+
+import { GdprServiceConnectionsRoot } from '../graphql/typings';
+import useAuthorizationCode from './useAuthorizationCode';
+import { getDeleteScopes, getQueryScopes } from './utils';
+
+const GDPR_SERVICE_CONNECTIONS = loader(
+  './graphql/GdprServiceConnectionsQuery.graphql'
+);
+
+export type LoadStatus = {
+  loading: boolean;
+  authorizationCode?: string | null;
+  complete: boolean;
+  error?: ApolloError;
+};
+
+type Options = {
+  requiredGdprScope: 'delete' | 'query';
+  deferredAction: string;
+  onError?: (e: ApolloError, status: LoadStatus) => void;
+  onCompleted?: (authorizationCode: string, status: LoadStatus) => void;
+};
+
+function useServiceConnectionsAuthorizationCode(
+  options: Options
+): [() => void, LoadStatus] {
+  // Using "useRef" instead of "useState", because "useState" is async
+  // When calling updateStatus and then onCompleted(status) the status object would be old version.
+  const loadStatusRef = useRef<LoadStatus>({
+    loading: false,
+    complete: false,
+  });
+  const updateStatus = (newProps: Partial<LoadStatus>) => {
+    loadStatusRef.current = { ...loadStatusRef.current, ...newProps };
+  };
+  const [getAuthorizationCode] = useAuthorizationCode(
+    options.deferredAction,
+    authorizationCode => {
+      if (!authorizationCode) {
+        const error = new ApolloError({
+          errorMessage: 'Authorization code not loaded',
+        });
+        updateStatus({
+          error,
+          loading: false,
+        });
+
+        if (options?.onError) {
+          options.onError(error, loadStatusRef.current);
+        }
+      } else {
+        updateStatus({
+          complete: true,
+          loading: false,
+          authorizationCode,
+        });
+        if (options?.onCompleted) {
+          options.onCompleted(authorizationCode, loadStatusRef.current);
+        }
+      }
+    }
+  );
+
+  const [getServiceConnections] = useLazyQuery<GdprServiceConnectionsRoot>(
+    GDPR_SERVICE_CONNECTIONS,
+    {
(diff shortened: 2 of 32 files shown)


src/gdprApi/useServiceConnectionsAuthorizationCode.ts
```diff
@@ -0,0 +1,108 @@
...
+  const loadStatusRef = useRef<LoadStatus>({
+    loading: false,
+    complete: false,
+  });
+  const updateStatus = (newProps: Partial<LoadStatus>) => {
+    loadStatusRef.current = { ...loadStatusRef.current, ...newProps };
+  };
+  const [getAuthorizationCode] = useAuthorizationCode(
+    options.deferredAction,
+    e => {
```
akikoskinen: Name `e` usually refers to an error, but here it seems to be the authorization code. Would be good to rename the variable to better tell what it is.

src/common/expandingPanel/ExpandingPanel.tsx
```diff
@@ -41,9 +46,16 @@ function ExpandingPanel({
...
-  const { isOpen, buttonProps, contentProps } = useAccordion({
+  const { isOpen, buttonProps, contentProps, openAccordion } = useAccordion({
     initiallyOpen,
   });
+
+  useLayoutEffect(() => {
+    if (initiallyOpen && isOpen !== initiallyOpen) {
+      openAccordion();
+    }
+  }, [initiallyOpen, openAccordion, isOpen]);
```
akikoskinen, line 57: With this logical change, isn't the `initiallyOpen` property now basically just a controller for whether the panel is open or not? So should it be renamed to e.g. `open` or `isOpen`?

src/i18n/en.json
```diff
@@ -210,7 +212,14 @@
     "empty": "You currently don't have any services connected to your profile.",
     "explanation": "You have granted the following services the right to use the data stored on your Helsinki profile.",
     "servicePersonalData": "Information which the service will know about you:",
-    "title": "Services with permission to use your data"
+    "title": "Services with permission to use your data",
+    "deleteConnection": "If you want to delete your data from this service, <deleteButton>click here</deleteButton>.",
+    "connectionRemovalVerification": "By continuing, you will lose all the data stored in the {{serviceName}} service and will no longer be able to use the service until you re-register to the service.",
+    "connectionRemovalSuccess": "All the data we had stored in the {{serviceName}} service has been deleted successfully!",
+    "connectionRemovalForbidden": "We cannot delete data from these services, for example due to the transaction status or storage time.",
```
akikoskinen: Should be singular instead of plural (`connectionRemovalForbidden`)?

src/i18n/sv.json
```diff
@@ -210,7 +212,14 @@
     "empty": "Du har för närvarande inga tjänster anslutna till din profil.",
     "explanation": "Du har gett följande tjänster rätt att använda uppgifterna som sparats i Helsingforsprofilen.",
     "servicePersonalData": "Information som tjänsten vet om dig:",
-    "title": "Tjänster som har tillstånd att använda dina uppgifter"
+    "title": "Tjänster som har tillstånd att använda dina uppgifter",
+    "deleteConnection": "Om du vill ta bort dina uppgifter från denna tjänst, <deleteButton>tryck här</deleteButton>.",
+    "connectionRemovalVerification": "Om du fortsätter, förlorar du alla uppgifter som du sparat i tjänsten {{serviceName}} och du kan inte längre använda tjänsten förrän du registrerar dig i tjänsten på nytt.",
+    "connectionRemovalSuccess": "Alla uppgifter som vi sparat i tjänsten {{serviceName}} har tagits bort på ett lyckat sätt!",
+    "connectionRemovalForbidden": "Vi kan inte radera data från dessa tjänster, till exempel på grund av transaktionsstatus eller lagringstid.",
```
akikoskinen: Should be singular instead of plural (`connectionRemovalForbidden`)?

src/i18n/fi.json
```diff
@@ -43,7 +43,9 @@
     "delete": "Poista omat tiedot",
     "explanation": "Jatkamalla menetät kaikki tallennetut tietosi etkä voi enää käyttää seuraavia Helsingin kaupungin palveluita:",
     "noServiceExplanation": "Jatkamalla menetät kaikki tallennetut tietosi.",
-    "title": "Haluatko varmasti poistaa kaikki tietosi?"
+    "title": "Haluatko varmasti poistaa kaikki tietosi?",
+    "contactServiceToDelete": "Jos haluat silti poistaa tiedot, ole suoraan yhteydessä <linkToExternalServiceList>kyseiseen palveluun</linkToExternalServiceList> ja yritä profiilin poistoa sen jälkeen uudestaan.",
+    "urlToServiceList": "https://www.hel.fi/helsinki/fi/kaupunki-ja-hallinto/hallinto/palvelut"
```
akikoskinen, line 48: Shouldn't these texts (or at least `contactServiceToDelete`) logically be in the `serviceConnections` section like the rest of the texts in this PR?

Also this mentions trying to remove the **profile** again, while the user only tried to remove a single service connection. The original text in the design says "Jos haluat silti poistaa tiedot, ole suoraan yhteydessä kyseiseen palveluun ja yritä tietojen poistoa sen jälkeen uudestaan". This applies to all languages.

src/profile/components/serviceConnections/DeleteServiceConnectionModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const cannotDelete = status === STATUS_DELETE_FORBIDDEN;
+  const hasError = status === STATUS_ERROR || cannotDelete;
+  const isLoading = status === STATUS_LOADING;
+  const isDone = status === STATUS_DONE;
+  const isFinished = isDone || hasError;
+  const shouldShowModal =
+    status !== STATUS_NONE &&
+    status !== STATUS_ACKNOWLEDGED &&
+    status !== STATUS_CLOSED;
```
akikoskinen: This would be easier to read if it listed the statuses when the modal _should_ be shown. Now the reader has to look for the whole set of status values and remove these three from them, to understand when the modal is shown.

src/profile/components/serviceConnections/ServiceConnection.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (isDone) {
+      setDeletionStatus(STATUS_ACKNOWLEDGED);
+      onDeletion(service);
+      return;
+    }
+    if (hasError) {
+      setDeletionStatus(STATUS_CLOSED);
+      return;
+    }
+    setDeletionStatus(STATUS_CLOSED);
```
akikoskinen: The `if (hasError)` block seems to do exactly the same as the rest of the function. Can it be removed?

akikoskinen (changes requested):

src/common/expandingPanel/ExpandingPanel.tsx
NikoHelle: Almost, the panel cannot be closed from outside. If it is open, initiallyOpen === false, won't close it.

The internal state (close/open) of the ExpandingPanel should not be "exported" outside, or its state should be only outside. Making the component obsolete.

One way to do this would be to not render the Expanding panel at all, before the initiallyOpen is resolved. Problem is, there are other ExpandingPanels rendered and there would be flickering when the one to be deleted might be rendered later.

src/i18n/fi.json
NikoHelle: Good catch, the "contactServiceToDelete" must copied under  "serviceConnections " and changed. "urlToServiceList" is added and used also in profile deletion ([PR](https://github.com/City-of-Helsinki/open-city-profile-ui/pull/278))

src/i18n/sv.json
NikoHelle: Yes, these must be returned to translator.

src/i18n/en.json
NikoHelle: Yes, these must be returned to translator.

src/common/expandingPanel/ExpandingPanel.tsx
akikoskinen: Makes sense. So this new functionality could be described perhaps something like this: "for an already rendered ExpandingPanel, if it's currently closed, I want it to open (or if it's already open, nothing should happen)", right? From outside of the component, that sounds something like an action or an event, more than state. Is there a way to send "events" to components somehow? Something equivalent to calling a function perhaps?

NikoHelle: Yes, good description! It is a "force panel open" functionality. Just because the data, whether it should be open or not, is not available on first render.

Parent component has no access to child components functions. Child component does not exist, when parent is rendered. Profile UI uses functional components, so they have no instances to access. Usually parent passes callbacks to children and child can call those passed functions.  Child could call a passed function "shouldIBeOpen", but that is very unusual way to use React.

Javascript has events of course, but React does not use them for component-component communication. Except with Redux which has events. Kind of. It dispatches actions to state handlers. Parent and child can also share state and functions via React Context. But Redux and Context requires a shared data container, which is an overkill in this case.

Of course a window event listener could be added to the panel, and parent could emit an event. But that is not a common approach in React.

One solution would be to give React component a `key` which changes when the `initiallyOpen` changes. When component's `key`-prop changes, the old version is destroyed and a completely new instance is created with new state. But that solution has side-effects, because the panel is also closed when `initiallyOpen` changes. Unless a "openedOnce" is stored in the parent. :D But the Panel itself has that data, so it is left to that component to open itself, if `initiallyOpen` changes to `true`.

React has annoying quirks.

src/profile/components/serviceConnections/DeleteServiceConnectionModal.tsx
NikoHelle: I liked this more than having 5 checks of "===". But as the constant is `shouldShowModal` and not `shouldHideModal` it is more logical to use "===". This is now changed.

src/common/expandingPanel/ExpandingPanel.tsx
akikoskinen: Got it. Seems like there is no better way to achieve this.

akikoskinen (approved):

PR #278, opened 2022-11-09
HP-1686 better delete profile error handling

NikoHelle:
Show list of services which prevent profile deletion.


src/i18n/en.json
```diff
@@ -37,21 +37,23 @@
...
+    "deleteFailed": "The deletion failed for some reason. Please try again after a while!"
   },
   "deleteProfileModal": {
     "delete": "Delete my information",
     "explanation": "By continuing you will lose all of your stored data and you can no longer use the following City of Helsinki services:",
     "noServiceExplanation": "By continuing you will lose all of your stored data.",
     "title": "Are you sure you want to delete your information?",
+    "notAllowed": "You cannot delete your profile because one of the services using it needs your information, for example, for invoicing.",
+    "deletionErrorTitle": "The deletion failed!",
+    "genericError": "The profile could not be deleted. Please try again later.",
```
kanguran: deleteFailed and genericError are pretty similar Error messages, do we really need to duplicate them ?

src/profile/components/deleteProfile/__tests__/DeleteProfile.test.tsx
```diff
@@ -174,4 +196,27 @@ describe('<DeleteProfile /> ', () => {
...
+      });
+    });
+  });
+  it(`When deletion returns a generic error, an error message is shown`, async () => {
+    await act(async () => {
+      const testTools = await initTests(1);
+      const { waitForElement } = testTools;
+      await proceedUIToDeletionConfimed(testTools);
+      await waitForElement(errorDescriptionSelector);
+      expect(mockHistoryPushListener).toHaveBeenCalledTimes(0);
```
kanguran, line 219: Add waitFor as in other places

      await waitFor(() => {
        expect(mockHistoryPushListener).toHaveBeenCalledTimes(0);
      });

.env.development
```diff
@@ -2,4 +2,5 @@ REACT_APP_OIDC_AUTHORITY="https://tunnistamo.dev.hel.ninja/"
 REACT_APP_OIDC_CLIENT_ID="https://api.hel.fi/auth/helsinkiprofile-ui-dev"
 REACT_APP_PROFILE_AUDIENCE="https://api.hel.fi/auth/helsinkiprofiledev"
 REACT_APP_PROFILE_GRAPHQL="https://profile-api.dev.hel.ninja/graphql/"
-REACT_APP_OIDC_SCOPE="openid profile https://api.hel.fi/auth/helsinkiprofiledev"
\ No newline at end of file
+REACT_APP_OIDC_SCOPE="openid profile https://api.hel.fi/auth/helsinkiprofiledev"
+REACT_APP_PROFILE_BE_GDPR_CLIENT_ID="https://api.hel.fi/auth/helsinkiprofiledev"
```
kanguran, line 6: Can't we use "https://api.hel.fi/auth/helsinkiprofiledev" as parameter here.

NikoHelle: What do you mean "as parameter"?

src/i18n/en.json
NikoHelle: Good catch. Removed the old one (genericError)

src/profile/components/deleteProfile/__tests__/DeleteProfile.test.tsx
NikoHelle: No need in this case. WaitFor is necessary when the test won't resolve immediately and need to wait for component update.

.env.development
NikoHelle: *.env could use this, BUT it won't work in Azure, so it was removed earlier.

REACT_APP_PROFILE_BE_GDPR_CLIENT_ID="https://api.hel.fi/auth/helsinkiprofiledev"
REACT_APP_OIDC_SCOPE="openid profile ${REACT_APP_PROFILE_BE_GDPR_CLIENT_ID}"

kanguran: ok, clear

kanguran (approved):
Looks good now

PR #284, opened 2023-01-23
HP-962: cookie consents

NikoHelle:
Cookie consents are asked in a modal which is shown on all pages - except there is also a cookie settings page where modal is not shown. The settings page has same content, so no need to show both. Modal is rendered again when user navigates to another page.

Matomo tracking is disabled when cookie consent is not given. Even anonymous, non-cookie tracking. There is currently no verification from GDPR-lawyers can users be tracked without consent even if cookies are not used. 

Matomo tracking is never enabled in non-production environments. To test Matomo tracking, modify `enableTrackingCookies()` in tracking.ts. Events should be pushed, so comment out the `return` in the if-clause.

### DIFF
--- a/src/cookieConsents/cookieContentSource.ts
+++ b/src/cookieConsents/cookieContentSource.ts
@@ -1,0 +1,117 @@
+import { ContentSource } from 'hds-react';
+import { TFunction } from 'react-i18next';
+
+export const trackingCookieId = 'matomo';
+
+export function getCookieContent(
+  t: TFunction
+): Pick<
+  ContentSource,
+  'requiredCookies' | 'optionalCookies' | 'texts' | 'siteName'
+> {
+  const securityCookieDescription = t('cookies.securityCookieDescription');
+  const loadBalancerCookieName = t('cookies.loadBalancerCookieName');
+  const loadBalancerCookieDescription = t(
+    'cookies.loadBalancerCookieDescription'
+  );
+  const languageCookieDescription = t('cookies.languageCookieDescription');
+  const trackingCookieDescription = t('cookies.trackingCookieDescription');
+  const consentStorageDescription = t('cookies.consentStorageDescription');
+  const sessionExpiration = t('cookies.sessionExpiration');
+  const daysPlural = t('cookies.daysPlural');
+  const minutesPlural = t('cookies.minutesPlural');
+  const mainText = t('cookies.mainText');
+  const siteName = t('appName');
+
+  const tunnistusHostName = 'tunnistus.hel.fi';
+  const profiiliHostName = 'profiili.hel.fi';
+  const tunnistamoAndProfileBeHostName = 'api.hel.fi';
+  const anyHelFiHostName = '*.hel.fi';
+
+  return {
+    siteName,
+    texts: {
+      sections: {
+        main: {
+          text: mainText,
+        },
+      },
+    },
+    requiredCookies: {
+      groups: [
+        { commonGroup: 'tunnistamoLogin' },
+        {
+          commonGroup: 'informationSecurity',
+          cookies: [
+            {
+              id: 'tunnistamo-csrftoken',
+              name: 'tunnistamo_prod-csrftoken',
+              hostName: tunnistamoAndProfileBeHostName,
+              description: securityCookieDescription,
+              expiration: `365 ${daysPlural}`,
+            },
+            {
+              id: 'profiili-csrftoken',
+              name: 'profiili-prod-csrftoken',
+              hostName: tunnistamoAndProfileBeHostName,
+              description: securityCookieDescription,
+              expiration: `365 ${daysPlural}`,
+            },
+          ],
+        },
+        {
+          commonGroup: 'language',
+          cookies: [
+            { commonCookie: 'keycloak-language' },
+            { commonCookie: 'suomifi-language' },
+            {
+              id: 'profile-language',
+              name: 'i18next',
+              hostName: profiiliHostName,
+              description: languageCookieDescription,
+              expiration: sessionExpiration,
+            },
+          ],
+        },
+        {
+          commonGroup: 'loadBalancing',
+          cookies: [
+            {
+              id: 'loadbalancer',
+              name: loadBalancerCookieName,
+              hostName: `${profiiliHostName}, ${tunnistamoAndProfileBeHostName}, ${tunnistusHostName}`,
+              description: loadBalancerCookieDescription,
+              expiration: sessionExpiration,
+            },
+          ],
+        },
+      ],
+    },
+    optionalCookies: {
+      groups: [
+        {
+          commonGroup: 'statistics',
+          cookies: [
+            {
+              commonCookie: trackingCookieId,
+            },
+            {
+              id: 'matomo-session',
+              name: '_pk_ses*',
+              hostName: anyHelFiHostName,
+              description: trackingCookieDescription,
+              expiration: `30 ${minutesPlural}`,
+            },
+            {
+              id: 'matomo-cookie-consent',
+              name: 'mtm_.*',
+              hostName: anyHelFiHostName,
+              description: consentStorageDescription,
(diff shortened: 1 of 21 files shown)


src/common/helpers/tracking/tracking.ts
```diff
@@ -0,0 +1,86 @@
+import { createInstance } from '@datapunt/matomo-tracker-react';
+import { useMemo } from 'react';
+
+import { trackingCookieId } from '../../../cookieConsents/cookieContentSource';
+
+type PaqPushEvent = string[];
+type PaqPush = (event: PaqPushEvent) => void;
+type Paq = { push: PaqPush };
```
akikoskinen: Does "paq" mean anything? At least to me it doesn't. As a result, it's quite hard to grasp the idea of these types.

src/common/helpers/tracking/tracking.ts
```diff
@@ -0,0 +1,86 @@
...
+      window._paq.length = 0;
+    }
+
+    const matomo = createInstance({
+      urlBase: 'https://analytics.hel.ninja/',
+      siteId: 60,
+    });
+
+    if (existingTrackingEvents.length) {
+      existingTrackingEvents.forEach(event => window._paq.push(event));
```
akikoskinen: This seems to do the same as `pushEvent` above. Could that function be called instead?

src/common/helpers/tracking/tracking.ts
```diff
@@ -0,0 +1,86 @@
+import { createInstance } from '@datapunt/matomo-tracker-react';
```
akikoskinen: This file is very Matomo specific, so should it be named `matomo.ts` or `matomoTracking.ts` or something like that?

src/cookieConsents/cookieContentSource.ts
```diff
@@ -0,0 +1,118 @@
...
+  const trackingCookieDescription = t('cookies.trackingCookieDescription');
+  const consentStorageDescription = t('cookies.consentStorageDescription');
+  const sessionExpiration = t('cookies.sessionExpiration');
+  const daysPlural = t('cookies.daysPlural');
+  const minutesPlural = t('cookies.minutesPlural');
+  const mainText = t('cookies.mainText');
+  const siteName = t('appName');
+
+  const tunnistusHostName = 'tunnistus.hel.fi';
+  const tunnistamoHostName = 'tunnistamo.hel.fi';
```
akikoskinen: I don't know if it makes a difference here, but Tunnistamo isn't (and probably never will be) reachable via tunnistamo.hel.fi URL.

akikoskinen (commented):
Seems to work nicely. A few comments. I also noticed https://github.com/City-of-Helsinki/helsinki-design-system/issues/947 if there's something you can do about that.

src/common/helpers/tracking/tracking.ts
```diff
@@ -0,0 +1,86 @@
...
+    window._paq = [];
+  }
+  return (window._paq as unknown) as Paq;
+}
+
+function pushEvent(events: PaqPushEvent): void {
+  events.forEach(event => {
+    getTrackingObject().push([event]);
+  });
+}
```
akikoskinen: This is quite an obfuscated API to Matomo. `matomo-tracker-react` also provides a [pushInstruction](https://github.com/jonkoops/matomo-tracker/blob/main/packages/react/src/useMatomo.ts#L42) function that better tells its purpose. Could it be used instead?

src/common/helpers/tracking/tracking.ts
NikoHelle: The "window._paq" is an array where Google Analytics originally stored it's events. Matomo/Piwik started to use an object with same name. 

I tried to google what it stands for, but didn't find anything. :D I've used it for so many years, it is clear what is stands for and I have no better name for it. 

I'll change it anyway to `Tracker` or similar.

akikoskinen: See also [my other comment](https://github.com/City-of-Helsinki/open-city-profile-ui/pull/284#discussion_r1085417499), as that could make this a non-issue.

src/cookieConsents/cookieContentSource.ts
NikoHelle: Ah, changed to api.hel.fi

src/common/helpers/tracking/tracking.ts
NikoHelle: Matomo API is kind of obfuscated - the docs are bad. :)

I wish functions of the Matomo instance/hook would work! But I did try them first. They just don't work when tracking is disabled first with 'requireConsent'. At least I think that causes it. Using the `pushInstruction` did not enable tracking after that. In some cases the tracking didn't stop after opting out.

Tried also other ways to disable tracking on start, like using the configuration options in `createInstance`. Did not work. Tracking just won't start. No errors are shown.

Another problem for Matomo seems to be multiple enable/disable events. User can enable tracking and cookies, then go to cookie settings page and disable/enable them many times. That was also a problem to Matomo as the tracking did not start or did not end.

Finally I got all scenarios working by just using the `_paq` object and using the events found here:
https://developer.matomo.org/guides/tracking-consent

src/common/helpers/tracking/tracking.ts
NikoHelle: I named it `tracking`,  because that's what it does and if tracking service is changed, changes can be done in the same file and filename can be the same. :)

src/common/helpers/tracking/tracking.ts
NikoHelle: Changed to that

src/common/helpers/tracking/tracking.ts
NikoHelle: Explanation in that comment.

src/common/helpers/tracking/tracking.ts
akikoskinen: Hmm, that makes sense too; Matomo is an internal implementation detail.

src/common/helpers/tracking/tracking.ts
akikoskinen: Oh, that's unfortunate. And surprising, as [`pushInstruction`](https://github.com/jonkoops/matomo-tracker/blob/main/packages/js/src/MatomoTracker.ts#L353) seems to do exactly the same thing, append the commands to the `window._paq` array.

src/common/helpers/tracking/tracking.ts
akikoskinen: Then again, Matomo isn't really hidden within this module. `useTrackingInstance()` returns a `MatomoTracker` which is to be used with a `MatomoProvider`, so this is quite Matomo specific looking from the outside also.

src/common/helpers/tracking/tracking.ts
NikoHelle: Yes. There are some differences: Matomo instance exists and `window._paq` is not an array anymore.

`window._paq` can be created manually and when Matomo (or Google Analytics) instance is created, it usually checks the `window._paq` for pre-existing events and overwrites the `_paq` with its own version and then appends pre-existing events there.

For some reason the `matomo-tracker-react` does not load `matomo.js` at all if `_paq` has events. Found that earlier in the source code, while wondering why pre-existing events do nothing.

BTW the `pushInstruction` is provided by a hook, and `pushEvents` are not always called inside a component - only place where hooks can be used.

NikoHelle:
In short: this PR would have been a lot easier if Matomo supported simpler consent storing/revoking. Plan B was to store the consents to a new util and command Matomo from there.

src/common/helpers/tracking/tracking.ts
NikoHelle: These are now changed

src/common/helpers/tracking/tracking.ts
NikoHelle: Changed to matomoTracking

akikoskinen (approved):

PR #286, opened 2023-02-08
HP-1734 end idle user's session

NikoHelle:
Tokens are not auto-renewed anymore. They are renewed when they are expiring and user has been active.

If expiring and user is not active, tokens are renewed when user is active again.

If user is idle for 10 minutes, user is logged out. A warning is shown after 8 minutes of inactivity.

To test faster:
- authService.ts: uncomment line 52 with "accessTokenExpiringNotificationTime"
- useManualTokenRenewal.ts: divide all intervals by 60 on lines 31-33


NikoHelle:
Implementation will change, so closing this PR.

PR #287, opened 2023-03-02
HP-1796: UI facelift

NikoHelle:
Changed styles according to the [new design](https://share.goabstract.com/ef4b8648-6b86-41b9-b1c2-f0c78721a436?collectionLayerId&mode).

Also cleaned up the codebase. Removed "multi-item"-related components, because those have been obsolete for a while. A lot of code was merged to make components simpler and after that it was easier to make style changes, because container elements were unified and common styles easier to use.

This branch is currently in https://profiili.dev.hel.ninja/


akikoskinen (commented):
Overall looks nice. I see some differences between the implementation and the design. Are these intentional, or should I point them out?

Also this one error view could perhaps use some style tuning:

Screenshot_20230306_113507

NikoHelle:
I'll have to make some changes to the buttons. Otherwise implementation should be ok, at least designer did not say anything. The design is based on hel.fi, but we use HDS, so they cannot match 100%.

I'll add padding to the error too.

akikoskinen (changes requested):
Not sure if this PR is the reason for this behavior, but I noticed the following. Locally I shut down Profile API, so that Profile UI can't save any profile data changes. When I edit some data and try to save, an error message is shown correctly. When I then click the save button again, Profile UI doesn't even try to send the data to Profile API, it just informs that saving was successful.

akikoskinen (commented):
Here's a collection of some error messages. While they all look nice on their own, they all look different. Would it be important to use a unified style for the error messages, or is that not a goal?

Screenshot_20230308_083047
Screenshot_20230308_083108
Screenshot_20230308_083331
Screenshot_20230308_083406
Screenshot_20230308_083452

NikoHelle:
Great summary of errors! Thx!

Here are fixed ones

image

And success messages
image

NikoHelle:
> Not sure if this PR is the reason for this behavior, but I noticed the following. Locally I shut down Profile API, so that Profile UI can't save any profile data changes. When I edit some data and try to save, an error message is shown correctly. When I then click the save button again, Profile UI doesn't even try to send the data to Profile API, it just informs that saving was successful.

This is also in the old version, produceable in the test env. But only with emails, where data is not sent, if new email matched the old one. Same feature was added to other fields in this pr. The bug is fixed now and tests were added to prevent it.

src/profile/components/phoneNumberEditor/PhoneNumberFormAndData.tsx
```diff
@@ -0,0 +1,264 @@
...
+    hasData,
+    testId,
+    isEditing,
+    currentAction,
+    actionHandler,
+    isNew,
+    editButtonId,
+    removeButtonId,
+    dataType,
+    noticationContent,
```
akikoskinen: Typo: `noticationContent`.

akikoskinen (changes requested):
Only one commit squashing and a typo left to fix.

The commit squashing was [commented into the extra commit](https://github.com/City-of-Helsinki/open-city-profile-ui/commit/794543061894899cb4690b85bfa39261397502ee).

src/profile/components/phoneNumberEditor/PhoneNumberFormAndData.tsx
NikoHelle: Fixed - also in all files where used

NikoHelle:
SonarCloud complained again about duplicate code in test files. Ended up re-writing all editing tests.

Gathered all test scenarios to one file and use that is all editing tests. 

Editing tests follow same flow in all tests ,so it was quite convenient to gather them. 

Data and input validation and test initialisations differ, so that took same time to convert to common tests. 

Additions start at in [this commit](https://github.com/City-of-Helsinki/open-city-profile-ui/pull/287/commits/e2bb6b844ba2f9460fe2e65386cc3ad56701641c)

akikoskinen (approved):
:+1:

PR #301, opened 2023-10-20
HP-2023: Get two authorization codes when needed

NikoHelle:
Sometimes GDPR-API actions need also authorization codes from Keycloak. Sometimes only Tunnistamo codes, sometimes only Keycloak, sometimes both.

A successful GDPR-API call requires:
- fetching user's service connections
- picking GDPR scopes from the results
- jump to Tunnistamo to get the auth code
- read the response
- make the actual GDPR-API call with the auth code

The process was split into multiple files without any clear logic where is the next jump. It would have been very confusing to add a new auth code fetcher to that callback hell.

So first I added a queue-system which runs async-actions until all are finished or one fails. Main "problem" is the required browser redirections to the OIDC servers. The queue state is stored to session storage, so the queue runner can continue when the browser returns.

All actions (get service connections, get GDPR scopes, etc) are split in to small, easy to read files which are listed in clear order. GDPR actions (download profile, delete profile, delete service connection) use the same actions to get the auth codes, then each have their own "main" action to execute the actual API call with the auth codes.

Fixes: https://helsinkisolutionoffice.atlassian.net/browse/HP-2023

This branch is deployed to https://profiili.dev.hel.ninja/login.  There is `dryRun:true` in the api calls, so nothing is actually deleted.


src/common/actionQueue/actionQueue.ts
```diff
@@ -0,0 +1,336 @@
...
+    filterQueue(completeFilter).length === queue.length;
+
+  const checkIfActionCanUpdate = (
+    action: Action,
+    newProps: Partial<ActionUpdateProps>
+  ) => {
+    if (completeFilter(action) || errorFilter(action) || !action.type) {
+      return false;
+    }
+    if (newProps.active && action.active) {
```
mikkojamG: I think that these to ifs could be combined:
 `if(
    completeFilter(action) ||
    errorFilter(action) ||
    !action.type ||
    ((newProps.active && action.active))
)`

or just return 
`
return (
 completeFilter(action) ||
 errorFilter(action) ||
 action.type ||
 (newProps.active && action.active)
)
`

src/common/actionQueue/actionQueue.ts
```diff
@@ -0,0 +1,336 @@
...
+    props
+  ) => {
+    const item = getByType(type);
+    if (!item) {
+      throw new Error(`Unable to update item. Item of type ${type} not found.`);
+    }
+    const newQueue = queue.map(action => {
+      if (action.type === type) {
+        return restore(item, { ...props, updatedAt: Date.now() });
+      } else {
```
mikkojamG: `else` seems unnecessary:


```
if (action.type === type) {
  return restore(item, { ...props, updatedAt: Date.now() });
}
return { ...action };
```

src/common/actionQueue/actionQueueRunner.ts
```diff
@@ -0,0 +1,265 @@
...
+      return 'invalid';
+    }
+    if (action.complete) {
+      return 'complete';
+    }
+    const active = queueController.getActive();
+    if (active) {
+      if (active.type === action.type) {
+        return pendingPromise ? 'pending' : 'active';
+      } else {
```
mikkojamG: No need for `else` 🤔

src/common/actionQueue/actionQueueRunner.ts
```diff
@@ -0,0 +1,265 @@
...
+      return genericErrorTypes.CANNOT_EXECUTE_ACTION_IS_COMPLETE;
+    }
+    if (actionStatus === 'in-queue') {
+      return genericErrorTypes.CANNOT_EXECUTE_ANOTHER_ACTION_IS_ACTIVE;
+    }
+    if (actionStatus === 'not-next') {
+      return genericErrorTypes.CANNOT_EXECUTE_ACTION_IS_NOT_NEXT;
+    }
+
+    return genericErrorTypes.CANNOT_EXECUTE_ACTION_IS_INVALID;
```
mikkojamG: Potential use case for `switch` ?

src/common/actionQueue/actionQueueStorage.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+export function getStoredQueueData(storageKey: string) {
+  const storedQueue = getStoredQueue(storageKey);
+  return storedQueue
```
mikkojamG: Could this be replaced with 
```
return (
    storedQueue &&
    storedQueue.reduce(...
```

mikkojamG (commented):
Looks pretty solid. With this amount of changes it's really difficult to get a grasp of this thing as a whole.. 😅 Noticed couple of potential enhancements but nothing major.

src/common/actionQueue/actionQueue.ts
NikoHelle: Sonar wont allow so many conditionals, max is 3.

But the last lines can be simplified to 
`return !(newProps.active && action.active);`

src/common/actionQueue/actionQueue.ts
NikoHelle: Changed!

src/common/actionQueue/actionQueueRunner.ts
NikoHelle: Changed!

src/common/actionQueue/actionQueueRunner.ts
NikoHelle: Yes, `switch` made it clearer

src/common/actionQueue/actionQueueStorage.ts
NikoHelle: Done!

akikoskinen (approved):
I have not checked the code, but I tried this in practice and it seemed to work.

mikkojamG (approved):

PR #303, opened 2023-12-21
HP-2126: no service connections fix with auth codes

NikoHelle:
If user has no service connections and no auth codes to get, then the browser redirection action is executed twice and will throw an error.

Added a check so an action with active state **and** options.idleWhenActive **and** a pending promise will return status 'pending'. Previously it returned 'next'.

Also added a timeout to the redirection catcher, if the url does not match the expected url. If the browser is not actually redirected and the url is set in an useEffect, then the catcher checks the url before it is set.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,8 @@
     ]
   },
   "devDependencies": {
+    "@sinonjs/fake-timers": "^8.1.0",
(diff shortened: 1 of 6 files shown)


akikoskinen (approved):
Tried this in practice and it seems to fix the issue.

PR #305, opened 2024-02-06
HP-2220 | Change default value for Helsinki tunnus amr claim

charn:
`helusername` is deprecated and the new keycloak authentication uses `helsinki_tunnus` in tunnistamo (which Helsinki profile uses).

Refs: HP-2220


voneiden (approved):
Alrighty

Riippi (approved):

PR #334, opened 2024-04-15
HP-2344 | feat: update max length of name fields

charn:
First and last name need to be limited to 150. Name information is also updated to keycloak by the backend. Updating longer values will work, but unfortunately backend will only support tokens with name fields limited to 150. So upon the next login the API token will not be accepted by the backend anymore since the `User` object will be updated based on the information in the token.

Nickname only allows 32 characters.

Refs: HP-2344


Riippi (approved):
👍

PR #435, opened 2025-04-09
fix: Empty keycloak code HP-2818

Riippi:
Instead of empty string use dummy when kc auth code is not needed. Works now similar way that it was when tunnistamo code was not needed. Maybe fixes issues on dev.


mikkojamG (approved):

charn (approved):
LGTM! :+1: :shipit:

terovirtanen:
**HELSINKI-PROFILE-UI branch is deployed to platta: https://helsinki-profile-pr435-ui.dev.hel.ninja** :rocket::rocket::rocket:

PR #456, opened 2025-06-13
Fix: Link icon HP-2909

Riippi:
HDS update caused profiili-ui custom Link component to break down with huge icon. 
To fix this the custom Link component that was build on top of a copy of old HDS Link was removed and HDS Link is used instead.
Also fixes playwright tests by closing the cookie banner before some tests that seem to fail because of it. 

https://helsinkisolutionoffice.atlassian.net/browse/HP-2909


terovirtanen:
**HELSINKI-PROFILE-UI branch is deployed to platta: https://helsinki-profile-pr456-ui.dev.hel.ninja** :rocket::rocket::rocket:

mikkojamG (approved):

PR #516, opened 2026-04-29
fix: update hds switch esrun to tsx KEH-281

Riippi:
Update HDS to 5.1 
switch esrun to tsx  
Should fix these dependabot issues:
https://github.com/City-of-Helsinki/open-city-profile-ui/security/dependabot/114
https://github.com/City-of-Helsinki/open-city-profile-ui/security/dependabot/107

Also added missing localisation string to the google sheets that was missed in some past PR. 
There is no change in localisations on this repo, but it was visible when running the translations script.

Refs: KEH-281


mikkojamG (approved):
👍 🦭

